PR #5, opened 2017-09-01
Skylab Readme Update

kbergin:
I've updated the skylab readme a little bit. The template readme is in the wiki https://github.com/HumanCellAtlas/skylab/wiki/Template-Pipeline-Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ Secondary analysis pipelines for the Human Cell Atlas.
 # Pipelines
 - [Smart-seq2](https://github.com/HumanCellAtlas/skylab/tree/master/smartseq2_single_sample) secondary analysis pipeline
 
-**Indicates that this pipeline is blessed**
+**Bold indicates that this pipeline is blessed**
 
 # How to run pipelines from skylab
 For now, use git clone `git clone git@github.com:HumanCellAtlas/skylab.git` and run the pipeline in cromwell.
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ For now, use git clone `git clone git@github.com:HumanCellAtlas/skylab.git` and 
 1. WDL and Cromwell [Documentation](https://software.broadinstitute.org/wdl/)
 2. Running WDLs in [Cromwell](https://software.broadinstitute.org/wdl/documentation/execution.php)
 
+**[TODO] Update this section with better instructions on how to easily try our pipelines.**
+
 # [TODO] Blessed HCA Pipelines in Methods Repositories
 - Agora
 - Dockstore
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
This looks good to me. 

The only thing I could think to add would be a note to include versioning and source for the genome index and annotation used. Although this is defined within the associated reference wdl, it could be nice to have it upfront. 

Counter-argument to this suggestion would be that it would be one more location containing metadata that would need to be updated if the reference changes.

kbergin:
@ambrosejcarr , was your comment in regards to the Analysis Pipeline template readme in the wiki (linked above), or the skylab readme changes in the attached PR? The information about the reference used in the WDL is in the template linked above, here: https://github.com/HumanCellAtlas/skylab/wiki/Template-Pipeline-Readme#genome-reference

Someday when our references are stored in the blue box it may also be handy to list the specific UUID:Version of the bundle we are using for our pipeline.

kbergin:
I've added to the template to explicity state to include versions

ambrosejcarr:
@kbergin You are correct, I need more coffee; I reviewed the template readme (it's good!). 

Generally, I like this readme as well. 

One concern I have is that I am not sure, given this README, that I would be able to run the smartseq2 pipeline. 

Part of that is the links to external documentation in lieu of instructions, which could be a bit frightening to the uninitiated. Since it's relatively simple, If I was a user I would love to have an example with full commands to run the pipeline on either (1) google compute or (2, probably better) local, using specific examples (wdl, json, and toy data) from the smartseq2 pipeline. It'd be the hello-world for our pipelines, and users could feel good about being able to just download and run it. 

Then the documentation links could serve as a reference for users who might want to dig in and play around with the internals.

kbergin:
@ambrosejcarr That's a good plan, I agree it'd be useful for the user to easily be able to run our pipelines. Let me think through this some more

kbergin:
In the current version of the readme I've stated the way to run is a "for now" because I'm thinking the easiest way for someone to try our pipeline may be to eventually have a public firecloud workspace with the example and some example data for someone to try it out or clone it to their own firecloud space and play with it on their data.. Once our pipeline is blessed and in the public methods repos.

ambrosejcarr:
Yea, it's not a trivial thing to incorporate. I'd advocate merging the changes in this PR with an additional todo for later regarding the run instructions.

kbergin:
Yes, I was thinking at least some more info and some lists of TODOs would be, at the least, a step forward compared to the initial readme we made. 

I'll add a TODO to improve instructions on running our pipelines

kbergin:
Back to you @ambrosejcarr

PR #6, opened 2017-09-07
Add WDL for cellranger count

mckinsel:
Add a WDL that performs the tasks of `cellranger count`, starting with chunking and read formatting through to barcode filtering.


10x/wdls/count.wdl
```diff
@@ -0,0 +1,1072 @@
...
+  Array[File] split_files
+  
+  String d = "$"
+  String l = "{"
+  String r = "}"
+
+  command {
+    set -x  
+    
+    # There's a lot here, but it's mostly the same thing repeated. All the input reads and trimmed_seq
```
ambrosejcarr: Given that it's mostly the same thing repeated, is there any way to improve this? could you define a function or subtask that gets called here to do this to reduce the amount of repeated code?

10x/wdls/count.wdl
```diff
@@ -0,0 +1,1072 @@
...
+    # files need to be moved to paths in $PWD. Also, we have to prepare a json list of lists that
+    # has all these new paths so we can pass it to martian. The jq ' |= .+ [ ]' syntax just appends 
+    # something to a list. Finally, we need to create a list of empty lists called read2s that's the
+    # same length as the list of reads.
+    idx=0
+    reads_arr_arg=[]
+    read2s_arr_arg=[]
+    for reads_arr in ${d}(echo '[${sep="," reads_output}]' | jq -c '.[]'); do
```
ambrosejcarr: I guess there is something nice and clean doing everything in bash but I sure do like `.append()` more than `jq ' |= .+ [ ]'`

10x/wdls/count.wdl
```diff
@@ -0,0 +1,1072 @@
...
+      chemistry_def=setup_chunks.chemistry_def, gem_groups=extract_reads_join.gem_groups, sample_id=sample_id
+  }
+
+  call filter_barcodes_main {
+    input: sample_id=sample_id, matrices_h5=count_genes_join.matrices_h5, barcode_whitelist=setup_chunks.barcode_whitelist,
+      gem_groups=extract_reads_join.gem_groups, chemistry_def=setup_chunks.chemistry_def,
+      raw_fastq_summary=extract_reads_join.summary, attach_bcs_summary=attach_bcs_and_umis_join.summary,
+      barcode_summary=attach_bcs_and_umis_join.barcode_summary
+  }
+}
```
ambrosejcarr: Just curious, how long does this take to run on cromwell vs running the native pipeline?

10x/wdls/count.wdl
```diff
@@ -0,0 +1,1072 @@
...
+#     out string barcode_whitelist,
+#     src py     "stages/common/setup_chunks",
+# )
+task setup_chunks {
+
+  String sample_id
+  File sample_def
+  Array[File] read_paths
+  
+  String chemistry_name = "auto"
```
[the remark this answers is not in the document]
mckinsel: In the new PR, I've switched to the heredoc syntax. That gets rid of dollar = "$", but I think we still have to define the braces.

10x/martian-cli/martian_cli/cli.py
```diff
@@ -166,55 +165,31 @@ def get_parser(stages):
                     help_message += " Help: " + input_.help
                 phase_parser.add_argument(
                     "--" + input_.name,
-                    type=martian_type_to_python_type(input_.type),
-                    nargs=martian_type_to_nargs(input_.type),
+                    type=_maybe_parse_json,
```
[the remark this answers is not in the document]
mckinsel: I've gotten rid of all this python code. We can use martian_shell.py now.

10x/wdls/example_count_input.json
```diff
@@ -0,0 +1,10 @@
+{
+  "count.read_paths": ["/path/to/fastqs"],
+  "count.sample_def": "sample_def.json",
+  "count.sample_id": "hm100",
+  "count.reads_per_file": 5000000,
+  "count.primers": [{"name": "P5", "seq": "AATGATACGGCGACCACCGAGATCT"}, {"name": "P7", "seq": "CAAGCAGAAGACGGCATACGAGAT"}, {"name": "R1", "seq": "ACACTCTTTCCCTACACGACG"}, {"name": "R2", "seq": "GTGACTGGAGTTCAGACGTGTG"}, {"name": "switch_oligo", "seq": "AAGCAGTGGTATCAACGCAGAGTACATGGG"}, {"name": "polyA", "seq": "AAAAAAAAAAAAAAAAAAAA"}],
```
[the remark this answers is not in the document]
mckinsel: Fixed in the new PR.

10x/wdls/count.wdl
```diff
@@ -0,0 +1,1072 @@
...
+  
+
+  runtime {
+    docker: "10x_hca_wdl:latest"
+  }
+
+  output {
+    File chunks = "chunks" 
+    File chemistry_def = "chemistry_def"
+    String barcode_whitelist = read_string("barcode_whitelist")
```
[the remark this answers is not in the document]
mckinsel: The python code was just json.dumping outputs to files with names from the MRO. In this case, that turned out to insert extra quotation marks.

10x/wdls/count.wdl
mckinsel: I like to think of it as having an austere beauty, but maybe a little writeup of jq tricks would help?

10x/wdls/count.wdl
mckinsel: It's pretty similar. There are a couple of differences. First, cromwell makes a new copy of the reference_path for every task that has it as an input. So when align_reads starts, you have to wait while it copies ~100GB of identical reference data. And, cromwell seems to want to maximize parallelization, so it will run all instances of STAR at once, even though each instance loads a private copy of the SA into memory.

10x/wdls/count.wdl
ambrosejcarr: Hah. :smile: Could you use a lambda to define it once and then reuse the defined function? That would clarify and reduce code clutter (?)

10x/wdls/count.wdl
ambrosejcarr: Interesting & good to know. Thanks!

10x/wdls/count.wdl
mckinsel: I'm not sure how to do that in bash in a way that doesn't feel very exotic. We could have some named functions, but I think we'd have to define those in every step. We could maybe add some simple executables in the docker image called something like `jq-append`. That would clean up the WDL a lot, though the tradeoff would be that you'd have to inspect the image to see what's really going on. What do you think?

mckinsel:
I'm going to close this PR without merging. There have been some changes from 10x that led to a significant redesign, so rather than cramming it all here it's in a separate PR: #12

10x/wdls/count.wdl
ambrosejcarr: I think I prefer the existing method. There's a bit more code but at least it's all there and the user can look it up if they want to know what it's doing
.

PR #7, opened 2017-09-08
Ss2 build reference

jishuxu:
main changes.
1. add demo json. single chrom chr21 and 10k fastq PE reads(extracted from aligned bam file)
2. create "BuildRef bundle" workflow. Generate index files for STAR and rsem.
3. Some changes in pipeline wdl. 
a) change reference package to use tar file instead of tar.gz
b) increase featureCounts task memory to 15G
c) switch to standard hg19 reference bundle. Previous KCO bundle generated warning in FeatureCounts task.
d) add a workflow configuration json which can turn off CallCaching


kbergin:
Can you add a WDL that generates that downsampled data in `smartseq2_single_sample/Demo/ss2_single_sample_demo.json` ?

TimothyTickle:
I think that the submission/ingest process makes all json files but the analysis.json but that one happens after analysis. Not sure reference bundle scripts would make those.

smartseq2_single_sample/ss2_single_sample.wdl
```diff
@@ -42,7 +42,7 @@ task FeatureCountsUniqueMapping {
   }
   runtime {
     docker:"humancellatlas/star_dev:v1"
```
ambrosejcarr: Do we have a standard tagging method for docker images? I'm looking and I see a mix of:
1. latest
2. v.major
3. v.major.minor x2

smartseq2_single_sample/BuildRef/BuildReferenceBundle.wdl
```diff
@@ -0,0 +1,63 @@
...
+workflow BuildRef {
+  File fasta
+  File gtf
+  
+  call BuildStarReference {
+    input:
+      ref_fasta = fasta,
+      gtf_file = gtf
+  }
+  call BuildRsemReference{
```
ambrosejcarr: Inconsistency in formatting: below the inputs are input=value, above they are input = value.

dshiga: I find the extra spaces a bit clearer, input = value. We need that linter.

kbergin:
I wanted to add a comment that I know I didn't clearly state having a WDL to create the testing data in the ticket or AC for the ticket for this work. I think we discussed it a little in a meeting, but I may be totally making that up. If you don't think you have the capacity this sprint to add in the WDL we can bump it into a new ticket! I'm sorry about that, I will work on being more detailed and clear in tickets and ACs.

kbergin:
@TimothyTickle Sorry my comment wasn't clear. I didn't mean that the WDL would generate the json file, I meant it should generate the actual data. Whatever steps were taken to downsample the fasta files in WDL form so that the output of the WDL is a smaller fasta file (in this case, the demo data).

ambrosejcarr:
Kylee, I think the tool I used for 10x could be adapted to work on this
data if I steal Jishu's STAR task. I'll talk to her on Monday about it.
Shouldn't take more than an hour to create and test if we want to go that
route.

jishuxu:
Split wdl to build reference for RSEM and STAR separately and some minor formatting changes

jishuxu:
I have a script in wdl to do down-sample and used to generated demo data. I can talk with Ambrose and see which one we want add into PR.

PR #8, opened 2017-09-08
Add wrapper wdl for smartseq2 single sample

dshiga:
Adds a wrapper wdl for the smartseq2 single sample wdl. The smartseq2 wdl itself expects fastqs as inputs. The wrapper takes a bundle uuid and version and calls a prepare and analyze wdl, which translates the bundle uuid and version into fastq paths and then runs the smartseq2 wdl. Then the the wrapper wdl calls a submission task, which is just a placeholder for now, to be expanded in a future PR. The secondary analysis service will run the wrapper wdl, which it needs because it is triggered by notifications that contain just a bundle uuid and version.

See https://elastc.com/c/Kl2fPT8w/177-smartseq2-wrapper-wdl and https://elastc.com/c/ver8mz-b/123-preprocessing-wdl-for-smartseq2-pipeline.

### DIFF
--- a/smartseq2_single_sample/hca-dcp/prepare_and_analyze_ss2_single_sample.wdl
+++ b/smartseq2_single_sample/hca-dcp/prepare_and_analyze_ss2_single_sample.wdl
@@ -1,0 +1,119 @@
+import "ss2_single_sample.wdl" as ss2
+
+task GetInputs {
+  String bundle_uuid
+  String bundle_version
+
+  command <<<
+    python <<CODE
+    import json
+    import requests
+    import subprocess
+
+    # Get bundle manifest
+    uuid = '${bundle_uuid}'
+    version = '${bundle_version}'
+    print('Getting bundle manifest for id {0}, version {1}'.format(uuid, version))
+
+    url = "https://hca-dss.czi.technology/v1/bundles/" + uuid + "?version=" + version + "&replica=gcp&directurls=true"
+    print('GET {0}'.format(url))
+    response = requests.get(url)
+    print('{0}'.format(response.status_code))
+    print('{0}'.format(response.text))
+    manifest = response.json()
+
+    bundle = manifest['bundle']
+    uuid_to_url = {}
+    name_to_meta = {}
+    for f in bundle['files']:
+        uuid_to_url[f['uuid']] = f['url']
+        name_to_meta[f['name']] = f
+    
+    print('Downloading assay.json')
+    assay_json_uuid = name_to_meta['assay.json']['uuid']
+    url = "https://hca-dss.czi.technology/v1/files/" + assay_json_uuid + "?replica=gcp" 
+    print('GET {0}'.format(url))
+    response = requests.get(url)
+    print('{0}'.format(response.status_code))
+    print('{0}'.format(response.text))
+    assay_json = response.json()
+    
+    print('Downloading sample.json')
+    sample_json_uuid = name_to_meta['sample.json']['uuid']
+    url = "https://hca-dss.czi.technology/v1/files/" + sample_json_uuid + "?replica=gcp" 
+    print('GET {0}'.format(url))
+    response = requests.get(url)
+    print('{0}'.format(response.status_code))
+    print('{0}'.format(response.text))
+    sample_json = response.json()
+    sample_id = sample_json['geo_sample']
+
+    fastq_1_name = None
+    fastq_2_name = None
+    for f in assay_json['files']:
+        if f['format'] == '.fastq.gz' and f['type'] == 'read1':
+            fastq_1_name = f['name']
+        if f['format'] == '.fastq.gz' and f['type'] == 'read2':
+            fastq_2_name = f['name']
+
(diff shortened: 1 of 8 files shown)


smartseq2_single_sample/hca-dcp/prepare_and_analyze_ss2_single_sample.wdl
```diff
@@ -0,0 +1,119 @@
...
+    manifest = response.json()
+
+    bundle = manifest['bundle']
+    uuid_to_url = {}
+    name_to_meta = {}
+    for f in bundle['files']:
+        uuid_to_url[f['uuid']] = f['url']
+        name_to_meta[f['name']] = f
+    
+    print('Downloading assay.json')
```
ambrosejcarr: I expect this kind of wdl will need to be replicated across pipelines, do you think it would be worth it to extract these pieces of code into a package, with a `download_from_uuid` function?

I was separately wondering if we could convert the python portion of the secondary-analysis repo into something installable so I could reuse your cromwell commands. That would be a good place to put a function like the above, which would then be called twice more below to get the other files. 

Alternatively, you could inline `def download_from_uuid` to save on code and improve later maintainability if we ever need to change. I don't know if inline function defs are too meta. 😄 

I'd be happy to contribute to any of this.

smartseq2_single_sample/hca-dcp/prepare_and_analyze_ss2_single_sample.wdl
```diff
@@ -0,0 +1,119 @@
...
+        f.write('fastq_1\tfastq_2\tsample_id\n')
+        f.write('{0}\t{1}\t{2}\n'.format(fastq_1_url, fastq_2_url, sample_id))
+    print('Wrote input map')
+    CODE
+  >>>
+  runtime {
+    docker: "humancellatlas/secondary-analysis-python"
+  }
+  output {
+    Object inputs = read_object("inputs.tsv")
```
ambrosejcarr: Aside: I didn't realize how `read_object` worked! this is cool. 😄

smartseq2_single_sample/hca-dcp/submit_ss2_single_sample.wdl
```diff
@@ -0,0 +1,37 @@
+task submit {
+  String i
+
+  command {
+    cat ${i} > out.txt
+    echo s >> out.txt
+  }
+  runtime {
+    docker: "python:2.7"
```
ambrosejcarr: is this docker image required here? I don't see any python code.

smartseq2_single_sample/hca-dcp/wrapper_ss2_single_sample.wdl
```diff
@@ -0,0 +1,40 @@
+import "prepare_and_analyze_ss2_single_sample.wdl" as prep_and_analyze
```
ambrosejcarr: I re-read the description you wrote in the PR and readme comment ("do I need this?") and I get a sense that there are multiple reasons services calling different wrappers. I wanted to confirm that the prep_and_analyze wdl is expected to be run in multiple different services, and is generally be useful by itself, outside of the context of this calling wrapper.  

If not, I'd prefer to have `prepare_and_analyze_ss3_single_sample.wdl` broken apart and brought directly into this wrapper (3 imports of workflows instead of 1 workflow and 1 workflow wrapper). This would reduce the number of levels and simplify the logic that needs to be followed to understand what's being called.

smartseq2_single_sample/hca-dcp/prepare_and_analyze_ss2_single_sample.wdl
dshiga: Good suggestions. I like the idea of extracting some of the Python from the WDL into a package. Made a ticket for it:
https://elastc.com/c/2BDXUEQ4/197-extract-wrapper-wdl-python-into-package

smartseq2_single_sample/hca-dcp/wrapper_ss2_single_sample.wdl
dshiga: I think that's a good suggestion. I initially thought that people might want to reuse that prepare and analyze wdl but now can't think of a good reason why someone would want to - they'd be running the same analysis on the same data as the secondary analysis service does. Best to keep it simple for now and add that layer at a later time if we find a strong use case.

smartseq2_single_sample/hca-dcp/submit_ss2_single_sample.wdl
dshiga: Sorry that is a bit confusing. Cromwell requires a docker image for each task. I usually use the python image as a default, just to have an image for the task. However, you have a good point - that can be confusing. I have switched it to ubuntu.

dshiga:
Okay, I think I've addressed all your comments now, with a new commit plus some new tickets. Let me know if this looks good to merge or if you have any other thoughts.

ambrosejcarr:
Looks ready for merge to me!

PR #9, opened 2017-09-08
Create 10x demo data

ambrosejcarr:
This PR addresses two tickets:

1. 10x demo data prep (`10x/generate_demo_data`)
2. Create a resource bundle for 10X V2 pipeline (`10x/generate_reference_bundle`)

It was not an AC for either ticket, but I feel like these two might benefit from a README file. I have not written one yet. If that's desirable I can write one up prior to the PR being reviewed (if it is, It would be great if someone could point me to an example that I should base mine off). 

In addition, as it stands now this wdl must be called once per sequencing lane. Since most experiments don't have many lanes, this is not very laborious (the demo data will have two lanes total). I can create a wrapper wdl to weave these calls together if desired.


ambrosejcarr:
@TimothyTickle Good catch on the runThreadN. I added a default parameter. I also pushed up some skeletons for READMEs for the two wdls. 

Speaking of defaults, I wanted to put default definitions in the type definitions as you suggest -- I find it a bit confusing to define it inline (what if I define two usages the same parameter with two different default values??). However, I don't see any documentation for support of what you're asking for in the [default category of the wdl spec](https://github.com/broadinstitute/wdl/blob/develop/SPEC.md#default).

@dshiga @kbergin do either of you know if what Tim is asking about is possible?

dshiga:
If you assign a value when declaring a variable, it is interpreted as a default value, which can be overridden by passing in another value when calling the task.

Int? subset_size = 100000

https://gatkforums.broadinstitute.org/wdl/discussion/10134/how-do-you-specify-default-values-in-a-wdl-runtime-block

The downside with this syntax is that it's not as obvious to the uninitiated that this is a default that can be overridden. (And WDL does not permit the "${default=foo bar}" syntax outside the command block.)

ambrosejcarr:
Ah, this is great. We should definitely use this syntax! I edited the WDL, verified that it works, and pushed a new commit reflecting these changes.

dshiga:
@TimothyTickle about pulling code out to make tools, I tend to think about it as follows, partly inspired by conversations with @ambrosejcarr:
- I don't mind fairly long bash or python scripts inline in tasks, because it's really convenient to have everything in one place.
- If there are tasks that we think will be used in multiple workflows, we should pull those tasks out into separate wdls that can be imported.

Factors I would consider when deciding whether to pull inline code out into a library:
- The longer and more complex it gets, the more we'll want to be able to break it out so we can make unit tests for it
- Some inline code is doing something pretty generic that could be useful in multiple tasks, in which case we probably want to pull it out into a library -- particularly if we think only certain pieces of the inline code are reusable, rather than the whole task (in which case we would move the task into its own reusable wdl)

PR #10, opened 2017-09-08
Test bare task and alt heredoc

ambrosejcarr:
**Don't merge this PR please, just for testing**

Tested two things: 

1. bare tasks seem to be importable
2. alternative heredoc syntax appears to allow differentiation within bash between bash variables (`$var`) and wdl variables (`${var}`)


kbergin:
This is cool!!

kbergin:
Something to keep in mind - Firecloud cannot run wdls that import other wdls. So if we want our analysis pipelines to be able to be easily ran in Firecloud in the near future, we will need to avoid using imports. I can push for that feature before production next year if Firecloud is approved as a blessed portal for HCA.

TimothyTickle:
Great catch Kylee, we definitely want the workflows to be runnable in FireCloud but that should not define best practices. As you were saying, we should use these pipelines to push best practice support in FireCloud, so coordinating with POs is an excellent suggestion. If FireCloud is or is not blessed, we will want these pipelines available for other projects in FireCloud.

ambrosejcarr:
Closing this; everyone has seen it and due to merges the PR is no longer informative (too many files changed vs master)

PR #13, opened 2017-09-14
Add wrapper for mock_smartseq2.wdl

samanehsan:
Add a wrapper wdl file for `mock_smartseq2.wdl` that takes in the bundle_uuid and bundle_version and passes the appropriate inputs to the mock_smartseq2 workflow.

### DIFF
--- a/mock_smartseq2/hca-dcp/prepare_and_analyze_mock_smartseq2.wdl
+++ b/mock_smartseq2/hca-dcp/prepare_and_analyze_mock_smartseq2.wdl
@@ -1,0 +1,101 @@
+import "mock_smartseq2.wdl" as ss2
+
+task GetInputs {
+  String bundle_uuid
+  String bundle_version
+
+  command <<<
+    python <<CODE
+    import json
+    import requests
+    import subprocess
+
+    # Get bundle manifest
+    uuid = '${bundle_uuid}'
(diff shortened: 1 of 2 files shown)


mock_smartseq2/hca-dcp/prepare_and_analyze_mock_smartseq2.wdl
```diff
@@ -0,0 +1,101 @@
...
+  }
+}
+
+workflow PrepareAndAnalyzeMockSmartseq2 {
+  String bundle_uuid
+  String bundle_version
+
+  File ref_tar
+
+  call GetInputs as prep {
```
ambrosejcarr: I have a syntax question I don't know the answer to: when we call a task as something else, technically tasks are supposed to be CamelCase. Should the "as" call mirror that here? 

`call GetInputs as Prep`
or
`call GetInputs as prep`

I think this code looks flawless. Nice job.

samanehsan: Just making a note here that we discussed this offline, and the conclusion was that we'll want to revisit this style question later and in general make sure we're being consistent with naming, capitalization, etc.

samanehsan:
@dshiga did you also want to take a look at this PR?

PR #14, opened 2017-09-14
Change barcode sort order to match cellranger

mckinsel:
This is a very small change to help make more outputs of the count
WDL bit-identical to the outputs from the cellranger tarball. The
pairs of bases always appear to show up in the same order when joining,
and though it really shouldn't make a difference, this appears to
affect the output if we use a different order.

### DIFF
--- a/10x/count/count.wdl
+++ b/10x/count/count.wdl
@@ -765,44 +765,44 @@ task sort_by_bc_join {
 
   command <<<
     set -x
-    
+
     echo '{}' > _args
-    echo '[{"prefix": "AA"}, ' >> _chunk_defs
-    echo '{"prefix": "AT"}, ' >> _chunk_defs
+    echo '[{"prefix": ""}, ' >> _chunk_defs
+    echo '{"prefix": "AA"}, ' >> _chunk_defs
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
Could you clarify what you mean by "the output" -- which output is affected, and how? I'm a bit concerned that changing something trivial like this could matter!

Related: your original sort order makes sense to me, cellranger's seems a bit random. Do you think there's anything to understand here? 

Generally I think this is fine to merge, but would love to hear your thoughts on these two questions. :)

mckinsel:
These are good questions, and I agree the behavior here is very mysterious. I was seeing a change in "num_alignments" in the _outs file of the sort_by_bc_join task. It seems impossible that this could matter though because the chunks get sorted by prefix anyway: https://github.com/10XGenomics/cellranger/blob/master/mro/stages/counter/sort_reads_by_bc/__init__.py#L80

mckinsel:
@ambrosejcarr I think you've been able to get good results not using this PR right? If so, it would be a relief to just close this and forget about it.

ambrosejcarr:
@mckinsel I am seeing identical results in the .mex and .h5 files, so I think that it's ok to close this.

PR #17, opened 2017-09-15
10x count gce

ambrosejcarr:
This is a pull request to modify `count.wdl` and its inputs to run on google compute engine. It produces identical output to vanilla `cellranger` on the same data. 

Still to-do is to explicitly specify the outputs. Of note, however, is that the outputs for each task are retained, and so the outputs are properly retained. This is a hash to successfully completed run: `f4841206-d251-4793-a6a7-554abadaf31d` where the outputs can be examined. 

This PR also confirms that the testing data and reference work with the count wdl.


10x/count/count.wdl
```diff
@@ -396,6 +405,8 @@ task align_reads_split {
   File reference_path
   
   command <<<
+    # todo is this necessary here? is the reference itself used or just the path to it?
+    mkdir genome_reference && tar -xf ${reference_path} -C genome_reference --strip-components 1
```
mckinsel: I believe all it's using is the reference.json in the reference path. I think there's going to be a bunch of optimizations like this where assumptions about a shared filesystem lead to inefficiencies when running on GCE. Probably shouldn't hold up merging this, but we should make a note for a future optimization ticket.

10x/count/count.wdl
```diff
@@ -78,16 +77,20 @@ task chunk_reads_split {
     /martian/adapters/python/martian_shell.py $stage_path $stage_phase . . run_file_whatever
     
     # Break the chunks into separate files so we can scatter over them
-    jq -c '.chunks[]' _stage_defs | split -l 1 - chunk_
+    jq -c '.chunks[]' _stage_defs | split -l 1 - partition_
```
mckinsel: I think we should use this naming for all of the split tasks. The choice of "chunk" in the first place was a little unfortunate because it's used in multiple other ways in the workflow.

10x/count/count.wdl
```diff
@@ -1126,17 +1188,23 @@ task filter_barcodes {
     stage_path=/cellranger/mro/stages/counter/filter_barcodes
     stage_phase=main
     /martian/adapters/python/martian_shell.py $stage_path $stage_phase . . run_file_whatever
+
+    # tar the filtered_matrices_mex folder so it can be extracted from the task
+    tar -cf filtered_matrices_mex.tar filtered_matrices_mex
```
mckinsel: Maybe compress this. I think the mtx file is stored in plain text and it probably compresses really well.

10x/count/count.wdl
ambrosejcarr: Great comment, and I agree with the sentiment. We shouldn't fix it now, but I made a ticket!

10x/count/count.wdl
ambrosejcarr: Fixed, good idea.

10x/count/count.wdl
```diff
@@ -8,20 +8,17 @@ task setup_chunks {
...
-    # Move reads files to the cwd so strings can point to cwd rather than the full
-    # cromwell path
-    mkdir read_path
-    for read_file in '${read_path}'/*; do
-      ln "$read_file" read_path/"$(basename $read_file)"
-    done
-     
+    # note that reads must be tar'd in a flat directory (tar -cf files.tar *) for
+    # this unpacking to function properly.
+    mkdir read_path && tar -xf ${read_path} -C read_path/
```
dshiga: What about giving this task an array of fastqs instead, and then having this task move them to read_path? I suspect that real input bundles will have individual fastqs and not tarballs.

10x/count/count.wdl
```diff
@@ -78,16 +77,20 @@ task chunk_reads_split {
...
-    jq -c '.chunks[]' _stage_defs | split -l 1 - chunk_
+    jq -c '.chunks[]' _stage_defs | split -l 1 - partition_
+
   >>>
 
   runtime {
     docker: "marcusczi/cellranger_clean:cromwell"
+    memory: "30 GB"
+    disks: "local-disk 50 HDD"
+
```
dshiga: Extraneous extra line

10x/count/count.wdl
```diff
@@ -103,12 +106,7 @@ task chunk_reads_main {
 
   command <<<
     
-    # Move the reads files to a path in the cwd. Note this needs to be a hard link
-    # Martian will deference soft links.
-    mkdir read_path
-    for read_file in '${read_path}'/*; do
-      ln "$read_file" read_path/"$(basename $read_file)"
-    done
+    mkdir read_path && tar -xf ${read_path} -C read_path/
```
dshiga: Hmm, File read_path is described above as a directory containing fastqs. As suggested in comment on setup_chunks, could we use an array of fastqs instead of a tarball here?

10x/count/count.wdl
```diff
@@ -404,7 +415,7 @@ task align_reads_split {
     read_groups_json='["${sep='","' read_groups}"]'
     echo "$(jq --arg var "$read_groups_json" '. + {"read_groups": ( $var | fromjson )}' _args)" > _args
     echo "$(jq --arg var "[]" '. + {"read2s": ( $var | fromjson )}' _args)" > _args
-    echo "$(jq --arg var "${reference_path}" '. + {"reference_path": $var}' _args)" > _args
+    echo "$(jq --arg var "genome_reference" '. + {"reference_path": $var}' _args)" > _args  # todo no required changes
```
dshiga: what does the todo mean? it sounds like it's a todo to not do anything :)

10x/count/count.wdl
```diff
@@ -436,11 +449,14 @@ task align_reads_main {
   File chunk
 
   command <<<
+
+    mkdir genome_reference && tar -xf ${reference_path} -C genome_reference --strip-components 1
+
```
dshiga: unnecessary empty line

ambrosejcarr: 👍

10x/count/count.wdl
ambrosejcarr: You're right; it was a note to myself that I didn't remove. Thanks for the catch.

10x/count/count.wdl
ambrosejcarr: 👍

10x/count/count.wdl
ambrosejcarr: On-point suggestion; This is something Marcus and I spoke about too. If we're getting these from cellranger mkfastq, I believe the input bundles should be directories with arbitrary numbers of fastqs (typically (1 to 8) * 3), so they're compatible with either approach. 

I elected to use the tarball because at the moment the pipeline processes these fastqs concurrently in a single wdl/cromwell call and because cellranger takes care of parsing the directory structure for us. 

As discussed this afternoon (I think just after this comment was made) there is an interest in breaking up earlier stages of the pipeline into single lanes and summarizing each lane with metrics. At that point it will *definitely* make sense to break this up into individual fastq files, as each wdl will take 3 fastqs (r1, r2, i1). 

I can definitely put in the change to fastq arrays; it will probably take 60-90 minutes to make the change + adjust the demo data and input files. Let me know if you think it's worth the time!

10x/count/count.wdl
ambrosejcarr: For now, I adjusted the comment to point out that there is a tarball here.

10x/count/count.wdl
mckinsel: I think the real type is `Array[Array[File]]` or like an `Array[Map[String, File]]`, though nesting of types seems to be disfavored by WDL.

But if this were to become an `Array[File]`, is there a preferred way to expand or iterate over that array in the command section? I keep using something like `'{sep="\' \'" var}'`, but maybe there's something better.

ambrosejcarr: I pushed a change here that uses this syntax:

```
  Array[File] r1
  Array[File] r2
  Array[File] i1
  # auto seems to be the only reasonable value
  String chemistry_name = "auto"

  command <<<

    # move all input files into read_path for cellranger
    mkdir read_path
    for fastq_file in ${sep=" " r1} ${sep=" " r2} ${sep=" " i1}; do
        mv $fastq_file read_path/
    done
```

mckinsel: That should work for this. But in general if the file is named `my file` or `* my_favorite_fastq *` or `; do sleep 1; done; sleep infinity` (whatever), I think it breaks. I'm not certain about that though.

Also one other thing to consider: the interface of this task suggests that the r1, r2, and i1 files are going to be grouped as they're ordered in the input arrays. So you'd expect something like a `for triplet in zip(r1, r2, i1)` somewhere. But it really just puts everything in the same directory and uses filenames to infer relationships between the files.

ambrosejcarr: Regarding the first point: yes, the user could give us some malicious code and this is not secure (unless cromwell checks to see that it is actually a file). 

I agree completely with your second point. We do make those expectations, and would expect to see a zip somewhere. However, the main goal was to change the inputs so that the wrapper wdls and other green box machinery which rely on 10x inputs can pass "good" inputs. 

The way those inputs are *used* is still bad, but the files that interface with the pipelines will now be reusable when we change the internals of this pipeline to be more robust. 

Does that make sense? I'm trying to minimize time spent making this pipeline more robust since the consensus seemed to be "replace everything".

mckinsel: Yeah, I'm mostly asking a general WDL question...so I can maybe do that in a venue other than this PR.

And I think it makes sense to delay some usability/robustness/etc discussions until we're somewhat closer to a workflow we're going to release.

PR #18, opened 2017-09-18
added output section

ambrosejcarr:
Aggregated outputs matching cellranger output from each task to create an outputs section in count.wdl. 

Note that the `summarize_reports` module of cellranger is not included in count.wdl, so we are missing the `metrics_summary.csv` and `web_summary.html` files that `cellranger` generates.


ambrosejcarr:
Seems like consensus from the meeting was that a selection of the current outputs from the 10x count wdl are acceptable, so I've made a PR.

PR #19, opened 2017-09-19
Submission wdl

dshiga:
Adds wdl for submitting an analysis bundle to ingest.

Adds the following:
- hca-dcp/submit.wdl - the submission wdl
- docker/secondary-analysis-python/tools/submit - Python package that does most of the submission work

I've modified the secondary-analysis-python Dockerfile to install the "submit" Python package.

The new submit wdl is generic and should be able to submit outputs from any analysis workflow.. Previously, we had a submission wdl that was specific to SmartSeq2.


docker/secondary-analysis-python/Dockerfile
```diff
@@ -5,3 +5,9 @@ LABEL software="Python"
 LABEL description="Python with some useful libraries"
 
 RUN pip install requests
+
+RUN mkdir /tools
+
+COPY tools /tools
```
ambrosejcarr: Could this be replaced by: 
```
RUN git clone https://github.com/HumanCellAtlas/skylab.git && \
    pip install -e skylab/docker/secondary-analysis-python/tools/submit
```

docker/secondary-analysis-python/tools/submit/setup.py
```diff
@@ -0,0 +1,18 @@
+from setuptools import setup
+
+setup(name='submit',
+      version='1.0.0.dev1',
+      description='submit',
```
ambrosejcarr: More informative description would be good.

docker/secondary-analysis-python/tools/submit/submit/submit.py
```diff
@@ -0,0 +1,83 @@
...
+    output_files = get_output_files(analysis_json_contents)
+    response = requests.put(file_refs_url, headers = json_header, data = output_files)
+    print(response.text)
+
+    # 6. Confirm submission
+    confirmation_url = '{0}/{1}'.format(submission_url, 'confirmation')
+    print('Confirming submission at {0}'.format(confirmation_url))
+    response = requests.put(confirmation_url, headers = json_header)
+    print(response)
+
```
ambrosejcarr: I think these calls could be a cool place to define a `notify` decorator, if someone wants to learn how those work at some point (low priority). It could look like this: 

```
from functools import wraps

def notify(notification_string):
    def decorator(func):
        @wraps(func)
        def wrapper(*args, **kwargs):
            print(notification_string)
            response = func(*args, **kwargs)
            print(response)
            return response
        return wrapper
    return decorator
```
Then, you can use this as follows: 

```
@notify('my_message_here')
def get_input_bundle_uuid(analysis_json):
    bundle = analysis_json['input_bundles'][0]
    uuid = bundle.split('/')[-2]
    print('Input bundle uuid {0}'.format(uuid))
    return uuid         
```
then you just invoke it with: 
```
input_bundle_uuid = get_input_bundle_uuid(analysis_json_contents)
```
and all the printing is already done. 

I doubt the code is runnable as I've coded it, but the idea is that you could abstract the notification and make it easy to write this. Flask may also already have this. :)

docker/secondary-analysis-python/tools/submit/tests/test.sh
```diff
@@ -0,0 +1,12 @@
...
+tag=$1
+if [ -z "$tag" ]; then
+  echo "Must pass in tag to use for docker image, e.g. test.sh test123"
+  exit 1
+fi
+cd ../../../
+docker build -t humancellatlas/secondary-analysis-python:$tag .
+cd -
+
+docker run humancellatlas/secondary-analysis-python:$tag bash -c "python -m unittest -v tests.test_submit"
```
ambrosejcarr: I really really like this testing method.

docker/secondary-analysis-python/tools/submit/submit/submit.py
ambrosejcarr: To make clear how decorators work: https://realpython.com/blog/python/primer-on-python-decorators/

docker/secondary-analysis-python/tools/submit/setup.py
```diff
@@ -0,0 +1,18 @@
...
+      url='http://github.com/HumanCellAtlas/skylab',
+      author='Dave Shiga',
+      author_email='dshiga@broadinstitute.org',
+      license='MIT',
+      packages=['submit', 'tests'],
+      install_requires=[
+          'requests',
+      ],
+      entry_points = {
```
ambrosejcarr: I had never seen or heard about this. I like this a lot. How did you come across this? I am interested because you must have searched for something totally different from my normal SO queries!

dshiga: I searched for "pip install callable module" and found some SO posts and blog posts about it.

ambrosejcarr: Very cool... thanks!

docker/secondary-analysis-python/Dockerfile
dshiga: The development process I want to enable is that I make changes to submit.py in my local copy of the repo, then I use the test.sh script to build and run the docker image and run unit tests in it. So for that use case it's important to have the docker image built from my local copy, so that my changes are reflected in it and tested.

I'm assuming that anyone else who has the secondary-analysis-python Dockerfile will have cloned the repo so that pip installing from tools/submit will work.

Does that make sense?

docker/secondary-analysis-python/tools/submit/setup.py
dshiga: Fair point, will add

ambrosejcarr:
Yep. Makes sense. Thanks!

docker/secondary-analysis-python/tools/submit/submit/submit.py
dshiga: I don't fully understand how this works, so would like to talk to you more about it.

docker/secondary-analysis-python/Dockerfile
ambrosejcarr: Makes perfect sense, thanks!

docker/secondary-analysis-python/tools/submit/submit/submit.py
ambrosejcarr: Probably true that it is overkill. It's a cool concept worth knowing about though. Happy to chat about it.

docker/secondary-analysis-python/tools/submit/submit/create_analysis_json.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python
+
+import requests
+import json
+import argparse
+from collections import OrderedDict
```
jishuxu: Do you use OrderedDict module? maybe analysis =OrderedDict() ?

docker/secondary-analysis-python/tools/submit/submit/stage.py
```diff
@@ -0,0 +1,134 @@
...
+    def _s3_chunk_size(file_size):
+        if file_size <= 10000 * 64 * MB:
+            return 64 * MB
+        else:
+            div = file_size // 10000
+            if div * 10000 < file_size:
+                div += 1
+            return ((div + (MB-1)) // MB) * MB
+
+def run():
```
ambrosejcarr: I can't see where this is used; was it for testing?

docker/secondary-analysis-python/tools/submit/submit/stage.py
```diff
@@ -0,0 +1,134 @@
...
+    runner = Main()
+
+if __name__ == '__main__':
+    runner = Main()
```
ambrosejcarr: convention is to use `_` or `unused` as variable names for classes whose return values are not used. 

Alternatively, you could just call `Main()` and not assign it.

docker/secondary-analysis-python/tools/submit/tests/test_get_staging_urn.py
```diff
@@ -0,0 +1,48 @@
+import unittest
+import os
```
ambrosejcarr: os, sys, json not used.

hca-dcp/submit.wdl
```diff
@@ -0,0 +1,163 @@
+# Get Cromwell metadata for the workflow that produced the given output
+task get_metadata {
+  String analysis_output_path
+
+  command <<<
+
+    # Get workflow_id
+    python <<CODE > workflow_id.txt
```
ambrosejcarr: I don't understand how this line is working. specifically the `> workflow_id.txt`

hca-dcp/submit.wdl
```diff
@@ -0,0 +1,163 @@
...
+
+    # Extract hash given a string like gs://foo/hash/call-bar/baz.txt
+    url = '${analysis_output_path}'
+    hash_end = url.rfind("/call-")
```
ambrosejcarr: I was going to suggest a regular expression here, but then couldn't be bothered to write it myself, so that's probably a sign that this is good enough 😬

hca-dcp/submit.wdl
```diff
@@ -0,0 +1,163 @@
...
+      -format_map ${format_map}
+
+    # Now create the submission object
+    create-envelope \
+      -submit_url ${submit_url} \
+      -analysis_json_path analysis.json
+  >>>
+
+  runtime {
+    docker: "humancellatlas/secondary-analysis-python:test16"
```
ambrosejcarr: I guess when you're ready to merge you could rename these dockers.

docker/secondary-analysis-python/tools/submit/submit/confirm_submission.py
```diff
@@ -0,0 +1,47 @@
...
+def get_envelope_json(envelope_url):
+    response = requests.get(envelope_url)
+    envelope_js = response.json()
+    return envelope_js
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-envelope_url', required=True)
+    parser.add_argument('-retry_seconds', type=int, required=True)
+    parser.add_argument('-timeout_seconds', type=int, required=True)
```
jishuxu: For retry_seconds and timeout_second. is it possible to give default values?  just curious

hca-dcp/submit.wdl
jishuxu: it is pipe to stdout :)

docker/secondary-analysis-python/tools/submit/README.rst
```diff
@@ -0,0 +1,81 @@
...
+
+The steps in the submission process are as follows:
+
+* Create analysis.json
+* Create submission envelope and upload metadata
+* Get URN need to stage files
+* Stage files
+* Confirm submission
+
+analysis_json.py
```
samanehsan: This is now `create_analysis_json.py`, right?

docker/secondary-analysis-python/tools/submit/README.rst
```diff
@@ -0,0 +1,81 @@
...
+================
+Creates analysis.json file.
+
+Invoke it like this::
+
+    analysis-json \
```
samanehsan: Same as above! It looks like this should be `create-analysis-json`

docker/secondary-analysis-python/tools/submit/README.rst
```diff
@@ -0,0 +1,81 @@
...
+      -run_type ${run_type} \
+      -method ${method} \
+      -schema_version ${schema_version} \
+      -inputs_file ${write_objects(inputs)} \
+      -outputs_file ${write_lines(outputs)} \
+      -format_map ${format_map}
+
+All arguments are required.
+
+submit.py
```
samanehsan: Is this now `create_envelope.py`?

hca-dcp/submit.wdl
```diff
@@ -0,0 +1,163 @@
...
+  String reference_bundle
+  String run_type
+  String schema_version
+  String method
+  Int retry_seconds
+  Int timeout_seconds
+
+  call get_metadata {
+    input:
+      analysis_output_path = outputs[0]
```
samanehsan: The `analysis_output_path` is referring to the `bam_file` from the analysis workflow, right? I think it would be clearer to pass it in as a separate input instead of making sure it's always the first item in `outputs`.

docker/secondary-analysis-python/tools/submit/submit/create_envelope.py
```diff
@@ -0,0 +1,108 @@
...
+    outputs = analysis_json['outputs']
+    output_refs = []
+    for o in outputs:
+        output_ref = {}
+        file_name = o['file_path'].split('/')[-1] 
+        output_ref['fileName'] = file_name
+        content = {}
+        content['name'] = file_name
+        content['format'] = o['format']
+        output_ref['content'] = content
```
samanehsan: I think this would be clearer to read if it was simplified to 
```
output_ref['content'] = {
    'name': file_name,
    'format': o['format']
}
```

docker/secondary-analysis-python/tools/submit/submit/get_staging_urn.py
```diff
@@ -0,0 +1,50 @@
...
+
+def get_staging_urn(envelope_js):
+    details = envelope_js.get('stagingDetails')
+    if not details:
+        return None
+    location = details.get('stagingAreaLocation')
+    if not location:
+        return None
+    urn = location.get('value')
+    if not urn:
```
samanehsan: You can remove this `if` statement, since `return urn` would just return `urn` or `None`

docker/secondary-analysis-python/tools/submit/submit/stage.py
```diff
@@ -0,0 +1,134 @@
...
+    print("\nPlease install boto3 to use this script, e.g. \"pip install boto3\"\n")
+    exit(1)
+
+KB = 1024
+MB = KB * KB
+
+
+def sizeof_fmt(num, suffix='B'):
+    """
+    From https://stackoverflow.com/a/1094933
```
samanehsan: A brief comment about what this function does would be helpful in addition to the stack overflow link!

samanehsan: Since this script is meant to be a temporary solution for the demo, it may make more sense to modify the original one instead.

docker/secondary-analysis-python/tools/submit/submit/stage.py
```diff
@@ -0,0 +1,134 @@
...
+        parser.add_argument('file_path', metavar="<file>",
+                            help="name of file to stage")
+        parser.add_argument('urn', metavar='<URN>',
+                            help="URN of staging area (given to you by Ingest Broker)")
+        self.args = parser.parse_args()
+
+    def _parse_urn(self, urn):
+        # TODO reimplement as deque
+        urnbits = urn.split(':')
+        if len(urnbits) == 6:
```
samanehsan: I'm not familiar with the format of an `urn`, or when one would have 5 items vs. 6, so adding a docstring with examples would be helpful here. This can go in as a later improvement to the original script though!

docker/secondary-analysis-python/tools/submit/tests/test_create_analysis_json.py
```diff
@@ -0,0 +1,84 @@
+import unittest
+import os
+import sys
+import json
+
+#pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
```
samanehsan: Were these comments meant to be removed?

docker/secondary-analysis-python/tools/submit/tests/test_create_analysis_json.py
```diff
@@ -0,0 +1,84 @@
+import unittest
+import os
+import sys
```
samanehsan: `sys` is not used here

docker/secondary-analysis-python/tools/submit/tests/test_create_analysis_json.py
```diff
@@ -0,0 +1,84 @@
...
+import json
+
+#pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
+#sys.path.insert(0, pkg_root)
+
+import submit.create_analysis_json as aj
```
samanehsan: I think using a more descriptive name, like `analysis_json` instead of `aj` would be more readable

docker/secondary-analysis-python/tools/submit/tests/test_create_analysis_json.py
```diff
@@ -0,0 +1,84 @@
...
+            for line in f:
+                output_lines.append(line.strip())
+        self.assertEqual(outputs[0]['file_path'], output_lines[0])
+        self.assertEqual(outputs[0]['format'], 'bam')
+        self.assertEqual(outputs[0]['name'], 'Aligned.sortedByCoord.out.bam')
+        self.assertEqual(outputs[1]['file_path'], output_lines[1])
+        self.assertEqual(outputs[1]['format'], 'metrics')
+        self.assertEqual(outputs[1]['name'], 'GSM1957573_rna_metrics')
+
+    def test_create_outputs(self):
```
samanehsan: This test has the same name as the one above it

docker/secondary-analysis-python/tools/submit/tests/test_create_analysis_json.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def test_get_format(self):
+        self.assertEqual(aj.get_format('asdf', {}), 'unknown')
+        self.assertEqual(aj.get_format('asdf.bam', {'.bam': 'bam'}), 'bam')
+        self.assertEqual(aj.get_format('asdf.txt', {'.bam': 'bam'}), 'unknown')
+        self.assertEqual(aj.get_format('asdf.bam', {'.bam': 'bam', '_metrics': 'metrics'}), 'bam')
+        self.assertEqual(aj.get_format('asdf.foo_metrics', {'.bam': 'bam', '_metrics': 'metrics'}), 'metrics')
+
+    def data_file(self, file_name):
+            return os.path.split(__file__)[0] + '/data/'  + file_name
```
samanehsan: This line looks like the indentation is off

docker/secondary-analysis-python/tools/submit/tests/test_create_envelope.py
```diff
@@ -0,0 +1,60 @@
+import unittest
+import os
+import sys
+import json
+
+#pkg_root = os.path.abspath(os.path.join(os.path.dirname(__file__), '..'))
```
samanehsan: Were these comments meant to be removed here as well?

docker/secondary-analysis-python/tools/submit/tests/test_create_envelope.py
```diff
@@ -0,0 +1,60 @@
...
+    def test_get_output_files(self):
+        with open(self.data_file('analysis.json')) as f:
+            js = json.load(f)
+            outputs = submit.get_output_files(js)
+            self.assertEqual(len(outputs), 3)
+            self.assertEqual(outputs[0]['fileName'], 'sample.bam')
+            self.assertEqual(outputs[0]['content']['name'], 'sample.bam')
+            self.assertEqual(outputs[0]['content']['format'], 'bam')
+
+    def test_check_status(self):
```
samanehsan: I think this test could be split up further based on the different error codes (e.g. `test_check_status_500`). I find having multiple specific tests over one test that does multiple things helps with readability and easier debugging.

samanehsan:
I had a few minor suggestions @dshiga but overall this looks really good! :octocat:

docker/secondary-analysis-python/tools/submit/submit/create_analysis_json.py
```diff
@@ -0,0 +1,135 @@
...
+        metadata = json.load(f)
+        start, end = get_start_end(metadata)
+        tasks = get_tasks(metadata)
+
+    inputs = create_inputs(inputs_file)
+    outputs = create_outputs(outputs_file, format_map)
+
+    input_bundles = get_input_bundles(input_bundles_string)
+
+    analysis = {}
```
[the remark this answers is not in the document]
dshiga: Good idea. On renaming things, these names are from the analysis.json spec so don't have a choice here.

docker/secondary-analysis-python/tools/submit/submit/stage.py
dshiga: I think you're right that it's unused. However, I copied this verbatim from https://github.com/HumanCellAtlas/staging-service/blob/master/scripts/stage_file.py and am reluctant to change it. Long term we should find a way to directly use the staging-service version of this file.

hca-dcp/submit.wdl
dshiga: What it means is "run the python script delimited by 'CODE' and then write the output to a file workflow_id.txt". It seems to be the right way to make this happen, though perhaps not the most intuitive syntax. I originally wanted to add a > after the ending 'CODE' but that didn't work.

docker/secondary-analysis-python/tools/submit/submit/confirm_submission.py
dshiga: Good idea, will do!

docker/secondary-analysis-python/tools/submit/submit/create_envelope.py
dshiga: That is better, thanks!

docker/secondary-analysis-python/tools/submit/submit/get_staging_urn.py
dshiga: Good point!

docker/secondary-analysis-python/tools/submit/submit/stage.py
dshiga: I copied that script verbatim from the staging service so don't actually know the answer and am reluctant to modify in any way:
https://github.com/HumanCellAtlas/staging-service/blob/master/scripts/stage_file.py

Long term we should find a way to use the script directly from the staging-service repo rather than having our own copy.

docker/secondary-analysis-python/tools/submit/tests/test_create_analysis_json.py
dshiga: Yeah, was an alternative implementation, good catch.

docker/secondary-analysis-python/tools/submit/tests/test_create_envelope.py
dshiga: Point taken - I split it into two, one for good and one for bad status codes.

hca-dcp/submit.wdl
dshiga: Hmm, let's talk about this in person. Basically we need any file output path, purely for the sake of parsing the workflow id from it.

dshiga:
I added 3 commits to this PR. @samanehsan @ambrosejcarr could at least one of you please take another look since you two had the most comments on the original PR? Please feel free to just click through to the diffs for the new commits rather than rereading everything.

- 51b89ab: This is a squashed version of Saman's PR here, which adds a wrapper for 10x: https://github.com/HumanCellAtlas/skylab/pull/25. I decided to base my branch on Saman's for expediency ahead of the demo.
- 9b86a70: Most of the changes are here, described further in commit message, stuff needed to make submission work for 10x and to make it more robust.
- cae3a2f: Changes to address the PR comments from 3 weeks ago.

dshiga:
@samanehsan it's because align_reads_main is called in a scatter but get_trimmed_and_aligned is not. So each align_reads_main task produces a File genome_output, but then outside the scatter block, align_reads_main.genome_output is an Array[File] that aggregates the files from each instance of the align_reads_main task. get_trimmed_and_aligned needs to consume the whole Array of Files, so I had to change it to expect genome_output to be an Array.

Really, both inputs, trimmed_seqs and genome_output need to be arrays, since they are zipped together inside the get_trimmed_and_aligned.

samanehsan:
Oh I see! Thanks for the clarification @dshiga

10x/hca-dcp/wrapper_10x_count.wdl
```diff
@@ -0,0 +1,340 @@
...
+      reads_per_file = reads_per_file,
+      subsample_rate = subsample_rate,
+      primers = primers,
+      align = align,
+      reference_path = reference_path,
+      umi_min_qual_threshold = umi_min_qual_threshold
+  }
+
+  String sample_id = GetInputs.inputs.sample_id
+  File primers_tsv = inputs_for_submit.primers_tsv
```
samanehsan: Where does this value get used?

10x/hca-dcp/wrapper_10x_count.wdl
```diff
@@ -0,0 +1,340 @@
...
+      ]
+  }
+
+  Array[Object] inputs = read_objects(inputs_for_submit.inputs)
+
+  call submit_wdl.submit {
+    input:
+      inputs = inputs,
+      outputs = [
+#        analysis.attach_bcs_and_umis_summary,
```
samanehsan: Should these outputs that are commented out be deleted?

10x/hca-dcp/wrapper_10x_count.wdl
dshiga: Apparently nowhere :) - thanks for catching that.

dshiga:
@ambrosejcarr about the python package being inside the docker dir: The docker build command doesn't let you use the COPY command to get files anywhere outside the directory where the Dockerfile lives. Since the purpose of this docker image is just to hold the secondary-analysis-python tools, it seems like a good idea to keep the Python code close to the Dockerfile anyway. We could think about renaming things so that the parent folder is not named docker but maybe tools. So we would have skylab/tools/submit, and inside submit there would be a Dockerfile as well as the Python package. That is probably best done in another ticket though.

docker/secondary-analysis-python/tools/submit/setup.py
```diff
@@ -0,0 +1,24 @@
+from setuptools import setup
+
+setup(name='submit',
```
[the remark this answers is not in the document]
dshiga: Yeah, I wasn't really thinking about other people installing this when I named it that. A rename might be in order depending on how we think people might use this package. Let's make another ticket for it.

dshiga:
This PR is no longer willing to update itself when I push commits. I will close and open a new one.

hca-dcp/submit.wdl
samanehsan: Ok, I see what you mean now!

PR #21, opened 2017-09-21
Refactor ss2 wrapper wdl and update getting workflow inputs

samanehsan:
- Refactor wrapper_ss2_single_sample.wdl to pass the inputs directly into the ss2 workflow instead of calling the prepare and analyze workflow.
- Update `GetInputs` task to get inputs from the new ss2 bundle


dshiga:
Could you please rename wrapper_ss2_single_sample_example_defaults.json? I called it defaults but that's not really accurate and might cause confusion - it's all the stuff that doesn't change from run to run. Maybe we could call it wrapper_ss2_single_sample_example_static.json.

PR #22, opened 2017-09-26
added monitoring script and options.json file

ambrosejcarr:
PR to add monitoring script + options.json that will: 
1. turn off call caching (otherwise the tasks might not run)
2. monitor the task's disk and memory usage every 5 seconds

Also added a bare-bones monitoring script to parse output monitoring logs so that this can be used in the short-term. 

These files will be wrapped by a framework to spawn, track, and parse multiple workflows with different input sizes. This work is still in progress.


monitoring/example_monitoring.log
```diff
@@ -0,0 +1,22 @@
+--- General Information ---
+#CPU: 1
+Total Memory: 1700
+Total Disk space: 10190136
+
```
jishuxu: I assume 1700Gb and 10190136kb disk?

jishuxu: never mind, I saw another log file, 1700MB

monitoring/monitor.sh
```diff
@@ -0,0 +1,23 @@
...
+        # print memory usage in megabytes
+        echo \* Memory usage \(MB\): $(free -m | grep Mem | awk '{ print $3 }')
+        # print disk usage as a percentage with up to 4 significant digits
+        echo \* Disk usage \(%\): $(df -k | grep cromwell_root | awk '{ OFMT="%.2f"; print ($3/$2)*100; }')%
+        # print disk usage in 1024 k blocks
+        echo \* Disk usage \(KB\): $(df -k | grep cromwell_root | awk '{ print $3; }')
+
+}
+
+while true; do runtimeInfo; sleep 5; done
```
jishuxu: checking usage in each 5s. How long does one 10X submission last normally?

monitoring/options.json
```diff
@@ -0,0 +1,5 @@
+{
+  "monitoring_script": "gs://broad-dsde-mint-dev-teststorage/10x/benchmark/scripts/monitor.sh",
+  "read_from_cache":false,
+  "write_to_cache":false
+}
```
jishuxu: That's neat. I did not know you can add script here.

monitoring/monitor.sh
jishuxu: It looks great. 'free' probably good enough for this purpose. alternatively, sometimes when I ran jobs on SGE, I used something like 'top -u username|grep command_name' to monitor jobs. it will automatically update and I pipe the output into a log file.

monitoring/parse_monitoring_log.py
```diff
@@ -0,0 +1,51 @@
...
+        max(disk_usage), total_disk, robust_disk_estimate,
+    )
+
+if __name__ == "__main__":
+    if any(arg in {'-h', '--help', 'h'} for arg in sys.argv):
+        print('usage: python3 parse_monitoring_log.py')
+    for file_ in sys.argv[1:]:
+        with open(file_, 'r') as f:
+            print('file %s' % file_)
+            max_mem, tot_mem, _, max_disk, tot_disk, _ = parse_task_log(f.readlines())
```
mckinsel: Does the monitoring script append a new resource utilization snapshot to a log file every 5 seconds? If so, you probably want to get rid of the `readlines` here, since I think that's going to try to read it all into memory.

Similarly, you might want to just store `max_memory_usage` and `max_disk_usage` rather than keeping the whole list around.

monitoring/monitor.sh
ambrosejcarr: Top was eating tons of CPU so I explicitly removed it from this monitoring script. I think top makes sense, but only if the interval between measurements is long.

monitoring/example_monitoring.log
ambrosejcarr: Good suggestion; I made this explicit.

monitoring/parse_monitoring_log.py
ambrosejcarr: Great suggestion; switched to iterators.

monitoring/parse_monitoring_log.py
```diff
@@ -0,0 +1,46 @@
...
+            total_memory = int(line.split()[-1])
+        elif line.startswith('Total Disk Space (KB):'):
+            total_disk = int(line.split()[-1])
+        elif line.startswith('* Memory usage (MB):'):
+            max_memory = max(max_memory, int(line.split()[-1]))
+        elif line.startswith('* Disk usage (KB):'):
+            max_disk = max(max_disk, int(line.split()[-1]))
+    robust_estimate = True if i >= 5 else False
+
+    return max_memory, total_memory, max_disk, total_disk, robust_estimate
```
samanehsan: Since this function has a lot of return values, I suggest returning these values in a dictionary so that you do not need to rely on their order. I don't expect this script will change much, but it's still a good practice :)

monitoring/monitor.sh
samanehsan: Would it be possible to run `parse_monitoring_log.py` on the `monitoring.log` here after this finishes?

monitoring/parse_monitoring_log.py
```diff
@@ -0,0 +1,46 @@
...
+      a task's cromwell directory within cromwell executions.
+    :return int max_memory: maximum memory usage within task
+    :return int total_memory: memory allocated for task
+    :return int max_disk: maximum disk usage within task
+    :return int total_disk: disk space allocated for task
+    :return bool robust_estimate: True if there were 5 or more estimates across the
+      task's run.
+
+    """
+    total_memory = None
```
samanehsan: Just wondering, is there a reason `total_memory` and `total_disk` are `None` and not `0`?

samanehsan:
I left a few comments on your PR @ambrosejcarr! 🌊

samanehsan:
Also, you should set your email in git so that your commits are associated with your GitHub account: https://help.github.com/articles/setting-your-commit-email-address-in-git/ :octocat:

monitoring/monitor.sh
ambrosejcarr: That's a great thought! 

Unfortunately, I think this script probably gets killed when the parent process completes, so there would be no guarantee that the script would run.

ambrosejcarr:
@samanehsan Great suggestions. Changes addressed; back to you.

ambrosejcarr:
@samanehsan I have added all of my emails to my github account; I'm not exactly sure how to check whether that fixed the problem you were highlighting. Please let me know!

monitoring/monitor.sh
samanehsan: Ok that makes sense!

samanehsan:
It's fixed! Your picture shows up next to your commits now :octocat:

PR #24, opened 2017-09-26
removed ds store

ambrosejcarr:
.DS_Store is a Mac operating system temp file. I don't *think* we need it in the repo.


mckinsel:
Should we add a .gitignore too?

PR #25, opened 2017-09-26
Add 10x count wrapper wdl

samanehsan:
Add a wrapper wdl that passes inputs to `count.wdl`, then sends the results to a submission wdl (which is currently a placeholder). The inputs that do not change with the bundle are defined in `wrapper_10x_count_example_static.json`, while bundle-specifc inputs are retrieved from google cloud storage.

Note this PR also includes a change to `count.wdl` to fix an error where Cromwell could not find the task `align_reads_main`


10x/count/count.wdl
```diff
@@ -1216,6 +1216,26 @@ task filter_barcodes {
...
+        memory: "30 GB"
+        disks: "local-disk 50 HDD"
+    }
+
+    output {
+      Array[Pair[File, File]] trimmed_and_aligned  =  zip(trimmed_seqs, genome_output)
+      Array[Pair[Int, Pair[File, File]]] gem_group_trimmed_and_aligned  =  zip(gem_groups, trimmed_and_aligned)
+    }
+}
+
```
mckinsel: Does moving this into a task mean that it's going to start up a new instance?

10x/hca-dcp/submit_10x_count.wdl
```diff
@@ -0,0 +1,37 @@
+task submit {
+  String i
+
+  command {
+    cat ${i} > out.txt
```
mckinsel: Does `${i}` refer to a local file?

10x/count/count.wdl
dshiga: Yes, it will. This is to work around a limitation in Cromwell when trying to run the count wdl in an import. It doesn't like having this zip command on its own outside of a task in that case. I suggested we put it in a task of its own to get around this, but I forgot that this would cause all the files to get copied down to one VM and that they might be big files. In general, will these be large files? If so, we should see if there's any way to work around this issue without having to copy everything to one VM.

mckinsel: They could be pretty big since believe this will download all the sequences and all their alignments. So it's probably something we'll have to address in the future. The way the align/extract tasks are parallelized is probably a little inconsistent with how WDL is supposed to work anyway.

What do you mean by "in an import"?

dshiga: So @samanehsan made a wrapper WDL that imports the 10x count WDL and takes a bundle id and version as inputs, talks to blue box to get the actual fastq paths, runs the 10x WDL, then submits the results back to ingest. We'll run it that way for the demo. In theory anything that works in a standalone WDL should also work when importing that WDL into another workflow, but that zip command seems to be a problem when importing only.

10x/hca-dcp/submit_10x_count.wdl
dshiga: Hmm, I think this is just a copy and paste from the placeholder submission WDL that I wrote for SS2, so this is my bad. Once this PR is merged we'll immediately replace this with a real submission WDL from this other PR:
https://github.com/HumanCellAtlas/skylab/blob/409ddb7501a74e4e86391fe4c7fb4fc70d19a45c/hca-dcp/submit.wdl
You're right though, we probably want to do echo instead of cat in this placeholder WDL.

10x/count/count.wdl
mckinsel: Ah okay, thanks for the explanation.

10x/hca-dcp/submit_10x_count.wdl
samanehsan: Yes! Thanks for pointing that out @mckinsel

samanehsan:
These commits are now part of #34

PR #27, opened 2017-10-05
Fix chunk sharding

mckinsel:
The `chunk_reads` steps are supposed to break up big fastq input files into little ones, using the `reads_per_file` parameter. But previously, if it actually did that, the pipeline would break. This change fixes the way that the chunked fastqs are distributed to subsequent workflow tasks.

### DIFF
--- a/10x/count/count.wdl
+++ b/10x/count/count.wdl
@@ -138,6 +138,30 @@ task chunk_reads_main {
     # Read the out_chunks definitions into a file
     jq '.out_chunks' _outs > out_chunks.json
 
+    # Here, out_chunks.json has a bunch of information about each of the fastq chunks that it's created,
+    # including the fastq files associated with each chunk. For 10x v2 chemistry, there are three files
+    # for each chunk, and they'll have names like R1-0001.fastq, R2-0001.fastq, etc. This command first
+    # gets all the file sets for each chunk by applying a compound filter to each element in the array
+    # in out_chunks.json. The filter first gets the value associates with the read_chunks key. Then it
+    # converts that JSON object into an array of objects with keys "key" and "value". Then it gets the
+    # "value" field in each of those and concatenates them with a space. So you start with something
+    # like
+    # [{"reads_chunks": {"R1": "path/to/R1_1.fastq", "R2": "path/to/R2_1.fastq", "II": "path/to/I1_1.fastq"}},
+    #  {"reads_chunks": {"R1": "path/to/R1_2.fastq", "R2": "path/to/R2_2.fastq", "I1": "path/to/I1_2.fastq"}}, ...]
+    # and you get a file like
+    # path/to/R1_1.fastq path/to/R2_1.fastq path/to/I1_1.fastq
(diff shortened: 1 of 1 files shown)


10x/count/count.wdl
```diff
@@ -137,6 +137,11 @@ task chunk_reads_main {
     
     # Read the out_chunks definitions into a file
     jq '.out_chunks' _outs > out_chunks.json
+    jq -r  'map(.read_chunks | to_entries | map(.value)) | map(join(" ")) | .[]' out_chunks.json > chunk_filenames.lines
```
ambrosejcarr: Sorry I still need a bit of hand-holding for code blocks like this.

10x/count/count.wdl
```diff
@@ -173,6 +182,17 @@ task chunk_reads_join {
     
     # Write out_chunks to their own file
     jq '.out_chunks' _outs > out_chunks.json
+
```
ambrosejcarr: ... and also some explanation here.

10x/count/count.wdl
```diff
@@ -161,7 +161,7 @@ task chunk_reads_join {
     
     # Create the args
     echo "{}" > _args
-    ehco "{}" > _chunk_defs
+    echo "{}" > _chunk_defs
```
ambrosejcarr: I'm sad to see `ehco` go.

10x/count/count.wdl
mckinsel: No problem, added some explanation. I'm basically just tarring the fastq chunks, but I have to parse the JSON file to know which files to include in which tarball.

10x/count/count.wdl
mckinsel: This is a very sad piece of code, but I added some comments. All I want to do is flatten a list of lists of files, but so far I seem to have to transfer everything to worker to do that.

10x/count/count.wdl
```diff
@@ -148,20 +172,24 @@ task chunk_reads_main {
...
-    Array[File] fastq_chunks = glob("files/*.fastq")
+    Array[File] fastq_chunks = glob("fastq_chunk_*.tar.gz")
   }
 }
 
 task chunk_reads_join {
   # Array of outs objects from the chunk_reads_main steps. All this step is doing
   # is joining some dicts together.
   Array[File] outs
+  Array[Array[File]] fastq_chunks
```
ambrosejcarr: I think for maintainability we should put some comments around inputs that aren't clear from their name/type that describe what they are. What you've got on lines 149 + 150 is good; we could copy it here. What do you think?

10x/count/count.wdl
samanehsan: I thought your explanation was really helpful! Thanks for adding it in @mckinsel

10x/count/count.wdl
mckinsel: Yeah good idea. I added a little note about what the `fastq_chunks` are supposed to be.

PR #29, opened 2017-10-09
Run SmartSeq2 automated test on DNAnexus

mckinsel:
This is not so much for merging as for some discussion. This is a PR that turns on a travis automated test of portability for the SmartSeq2 single sample WDL. It converts the WDL to a DNAnexus workflow using dxWDL, runs it, and checks the output.

The first commit has some changes that were needed to the WDL itself for it to work with dxWDL. dxWDL didn't like the implicit `String` to `File` conversion, and it didn't like the string interpolation happening in the workflow definition rather than in a task.

The second commit actually runs the test. Everything is kind of blended together in this proof-of-principle, but you can see some of the different concerns:
- "Compiling" the workflow definition for the external environment.
- Specifying the location of test inputs.
- Authentication to the external environment.
- I/O synchronization with the external environment.
- Checking the output produced by the external environment in a workflow-specific way.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,43 @@
+language: python
+
+python:
+    - "2.7"
+
+before_install:
+    - sudo apt-get -qq update
+    - sudo apt-get install -y openjdk-8-jre-headless
+
+install:
+    - wget https://wiki.dnanexus.com/images/files/dx-toolkit-current-ubuntu-14.04-amd64.tar.gz
+    - tar xf dx-toolkit-current-ubuntu-14.04-amd64.tar.gz 
+    - source dx-toolkit/environment
+    - wget https://github.com/dnanexus-rnd/dxWDL/releases/download/0.46/dxWDL-0.46.jar
+
+script:
+    - dx login --token "$DX_TOKEN" --noprojects && dx select "$DX_PROJECT"
+    - date_stamp="$(TZ=":US/Pacific" date +"%Y-%b-%d-%H-%M-%S")"
+    - dx mkdir "$date_stamp"
+    - workflow_id=$(java -jar dxWDL-0.46.jar compile smartseq2_single_sample/ss2_single_sample.wdl -destination "$DX_PROJECT":/"$date_stamp")
+    - >
+        analysis_id=$(dx run "$workflow_id" --yes --folder "$DX_PROJECT":/"$date_stamp" --brief -istage_0.fastq_read1=file-F7FxyyQ03bypfYvp1QPqXP7Y
+        -istage_0.fastq_read2=file-F7Fxyyj03byQ4F4z1Q7jkJ5F
+        -istage_0.gtf=file-F7Fxyy803byQ4F4z1Q7jkJ59
+        -istage_0.ref_fasta=file-F7Fxyx803byV6xvf1PZyzv5z
+        -istage_0.rrna_interval=file-F7Fxyy803bykQjFB1Pj3FV4b
+        -istage_0.ref_flat=file-F7Fxyz003byYVBkP1PzZ9B87
+        -istage_0.star_genome=file-F7FxyzQ03bypbvg11V4ZkXGJ
+        -istage_0.output_prefix="smartseq2_portability_test"
+        -istage_0.rsem_genome=file-F7Fxyz803byQ4F4z1Q7jkJ5J)
+    - while [ $(dx describe "$analysis_id" --json | jq  -r .state) = in_progress ]; do echo "I'm not dead!"; sleep 30; done
+    - dx wait "$analysis_id"
+    - metrics_hash=$(dx download $(dx describe "$analysis_id" --json | jq -r '.output."stage_6.aln_metrics"."$dnanexus_link"') -o - | tail -n +6 | md5sum | awk '{print $1}')
+    - test $metrics_hash = 3e42c0c7864e462173990f01afdb0a57
+
+env:
+    global:
+        - secure: "pkN4+Fm4+i8B5kLjlTccYr0/L6MeZfR9XoTN55Slu3hSmoG4xvc2QvhH4PcpJurKXbYNLYTSildfGHepounUsVsym5h2WjhikLj9bxVUwo/QlKhO2T175k8/H5fJn30ZG14cgjOwLjLtEmrY2OezPb87vFF/hxUhHerq+E6XkLjSLZtNrBQt34c7LfBWIollevaLFfGQel5zUjqcU7hZtmHDig3IxnMADXyYXTfsZf3vEV/PXsAmGnGY0U1PbUqacJTtJBrdJJiqjpLKhrtM+S7xv2KmcXqQqCeZ0693GsMYjl7qG5Iw7twH1muucOQKwJYY0ZjY/RBkZe/xTr8sBFFDhBxvrphoPjPJmJgTnbXdXC+XtpmpQpSQHAOYX1yathkJzb7FvYKmeeBRG8795rDwAz+7iuHBItdTYq/p3Rj7DIR4UUAzWMgiIfdbRI+T0M1FXJGpo7k63KGeUpqpS/ekWbW9qYPbTxGfDrR+qqz9RwTkZzs1WCIJl1lbmu0NFTbL32bbMhCdwTusK2CPlvNsQHX1j0GVKtOrbFYgFcUQEFLL0oweJyu1OK2Mt/gXpLlmc09Ljbv99EuAAsrH9kspu+HkP0+F+uWOXdpXSa/f8FsfzP3uchbJpabvcs0xVGbX2H+B//CnyhruTMoGMtB4MVMiycltVpSJpb41oTk="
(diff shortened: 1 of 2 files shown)


TimothyTickle:
Thank you Marcus,

This is an excellent exercise and hope we continue to do this and make infrastructure to automate this further. You are already showing need.

A quick question, is the required change moving the SS2 pipeline towards the current WDL spec or away? I think it is important that, at least, 1 of 3 actions should be evaluated:
1. Is our pipeline to spec? If so it should not change, if not it should be brought to spec.
2. Is the spec clear? If not, we should push it to be updated.
3. Is the execution environment that is evaluated conforming to the spec? If not, they should update to the spec.

This helps us optimized to the spec and clarify the spec as needed.

ambrosejcarr:
This is really cool; I don't have much to add about the content here, although regarding one of the changes, I think using `File` > `String` is a good practice when the object is a `File`. 

The Travis CI does seem to be running on my new branch as well, so it might be configured to run on all new branches. (click `branches`, you'll see my branch is failing). It hasn't stopped me from doing anything.

mckinsel:
Hey @TimothyTickle, I think there's no obvious way to answer these questions with the currently published WDL docs. For the type coercion, there's a table that shows that you should be able to coerce a `String` to a `File` when the string can be "Interpreted as file path". But it's not clear in what contexts that's allowed or what a valid "file path" is in every environment.

As far as the string interpolation, I think there's a major open question around what's allowed in a workflow definition vs a task definition vs a command section. The spec suggests that any declaration is allowed within the workflow definition, but all WDL executor implementations place limits on this. See a cromwell example in the discussion [here](https://github.com/HumanCellAtlas/skylab/pull/25#discussion_r141119226).

In any case, I think the couple changes I made are a move toward using "minimal" WDL features to accomplish the given task, which is a good goal for us to have.

mckinsel:
@ambrosejcarr, I turned on "Build only if travis.yml present" in the [settings](https://travis-ci.org/HumanCellAtlas/skylab/settings). So hopefully it won't try to build your branches anymore.

TimothyTickle:
Thank you for the clarification. I am looking forward to pushing for more clarity and, yes, minimal language tricks is good as long as we are not blocked to do our work (which you proven is not the case here :-) )

.travis.yml
```diff
@@ -0,0 +1,43 @@
+language: python
+
+python:
+    - "2.7"
```
dshiga: Python 3 would be preferable if there's nothing stopping us from using it.

smartseq2_single_sample/ss2_single_sample.wdl
```diff
@@ -135,9 +139,9 @@ workflow Ss2RsemSingleSample {
   File ref_fasta
   File rrna_interval
   File ref_flat
-  String star_genome
+  File star_genome
```
jishuxu: hm, can not recall why I used `String` instead of `File` here....looks like wdl can recognize a a `String` as a `File` if the `String` is pointed to an actually file?

.travis.yml
```diff
@@ -0,0 +1,43 @@
...
+
+before_install:
+    - sudo apt-get -qq update
+    - sudo apt-get install -y openjdk-8-jre-headless
+
+install:
+    - wget https://wiki.dnanexus.com/images/files/dx-toolkit-current-ubuntu-14.04-amd64.tar.gz
+    - tar xf dx-toolkit-current-ubuntu-14.04-amd64.tar.gz 
+    - source dx-toolkit/environment
```
dshiga: Do we definitely need dx-toolkit and its environment to run the jar? Might be preferable to install the minimum that we need with pip or apt-get: https://wiki.dnanexus.com/Downloads#Alternative-downloads. The tarball is a pretty opaque set of dependencies.

.travis.yml
```diff
@@ -0,0 +1,43 @@
...
+    - wget https://github.com/dnanexus-rnd/dxWDL/releases/download/0.46/dxWDL-0.46.jar
+
+script:
+    - dx login --token "$DX_TOKEN" --noprojects && dx select "$DX_PROJECT"
+    - date_stamp="$(TZ=":US/Pacific" date +"%Y-%b-%d-%H-%M-%S")"
+    - dx mkdir "$date_stamp"
+    - workflow_id=$(java -jar dxWDL-0.46.jar compile smartseq2_single_sample/ss2_single_sample.wdl -destination "$DX_PROJECT":/"$date_stamp")
+    - >
+        analysis_id=$(dx run "$workflow_id" --yes --folder "$DX_PROJECT":/"$date_stamp" --brief -istage_0.fastq_read1=file-F7FxyyQ03bypfYvp1QPqXP7Y
+        -istage_0.fastq_read2=file-F7Fxyyj03byQ4F4z1Q7jkJ5F
```
dshiga: Where did all these files come from? How do they get into the environment where Travis is running the job? Is it possible to give them more human readable names? Ideally we would gsutil cp these down from a standard location to make it clearer where they came from, but some comments in the yaml about their origin might do for now.

.travis.yml
```diff
@@ -0,0 +1,43 @@
...
+        -istage_0.gtf=file-F7Fxyy803byQ4F4z1Q7jkJ59
+        -istage_0.ref_fasta=file-F7Fxyx803byV6xvf1PZyzv5z
+        -istage_0.rrna_interval=file-F7Fxyy803bykQjFB1Pj3FV4b
+        -istage_0.ref_flat=file-F7Fxyz003byYVBkP1PzZ9B87
+        -istage_0.star_genome=file-F7FxyzQ03bypbvg11V4ZkXGJ
+        -istage_0.output_prefix="smartseq2_portability_test"
+        -istage_0.rsem_genome=file-F7Fxyz803byQ4F4z1Q7jkJ5J)
+    - while [ $(dx describe "$analysis_id" --json | jq  -r .state) = in_progress ]; do echo "I'm not dead!"; sleep 30; done
+    - dx wait "$analysis_id"
+    - metrics_hash=$(dx download $(dx describe "$analysis_id" --json | jq -r '.output."stage_6.aln_metrics"."$dnanexus_link"') -o - | tail -n +6 | md5sum | awk '{print $1}')
```
dshiga: Do I understand correctly here - we're downloading the alignment metrics file and doing an md5 on the non header lines? Some comments about what this and other steps are doing would be helpful.

.travis.yml
mckinsel: The only thing using python here is dx-toolkit, which sadly only works with python 2.7.

.travis.yml
mckinsel: These are just the files from the demo that have been uploaded to DNAnexus. It's definitely very opaque now. I think ideally we'd stage these files on DNAnexus from the blue box as part of the test. For now I'll add some comments, though.

.travis.yml
mckinsel: Yeah that's right, I'll add a comment. The idea here is to show a simple case of the concerns a "checker" tool would have to address.

smartseq2_single_sample/ss2_single_sample.wdl
mckinsel: Well cromwell/JES does in this case, but dxWDL/DNAnexus doesn't. There are cases where cromwell doesn't do the cast when I thought it would have, so I think the WDL type coercions tend to behave ambiguously. It's probably better to avoid relying on it unless we really need to.

.travis.yml
mckinsel: I think it might work with just pip installing `dxpy`; I'll test it. The docker point is interesting because we need to think about how to encapsulate the "stage" and "compile" behavior that's likely going to be required for running external tests. Putting it into a docker image might be the best solution for that.

mckinsel:
Hey @dshiga, thanks for taking a look! It does look like sub-workflows will cause issues. It seems like they're the most troublesome WDL feature, given that they don't work on Firecloud either (I think?). It's not clear to me why that is since, as you say, you could mostly implement them with a sed one-liner. I'm surely missing something, though. The Object thing is a little trickier, since WDL Objects are very lacking in documentation.

What do you think is a good process for hashing out the rough design of the portability tests? There is the big question of how to deal with differences between cromwell/JES and other environments. But there are lots of other design issues as well.

dshiga:
I'm told sub workflows are coming soon to FireCloud -- there was a pretty minor technical issue that was holding it back. I think a team meeting is probably a good next step and we can figure out from there if we should add to your portability doc, have further meetings, etc.

dshiga:
Following our discussion earlier this week, I think it makes sense to close this PR for now, with the understanding that in future we'll 1) build out further automated portability tests for other environments 2) not fail PRs due to portability issues (portability tests might or might not trigger off of PRs)

mckinsel:
Yep, I'll close this for now though I'll keep the branch around. I think our next steps are around two things:

1. Understanding the frequency and types of portability failures we should expect so we can figure out what to do about them.
2. Setting up some better infrastructure for test submission and querying. This example jams lots of concerns together, which won't be helpful for external implementers.

PR #30, opened 2017-10-09
Fix 10x Scaling Issues

mckinsel:
Previously, we were using command substitution and redirection to insert the contents of json files in the _args jsons. But when the contents of that file are really large, you get problems like "argument list too large". This PR replaces that with the slurpfile option from jq, which reads a json file and doesn't make you send its contents through the command line.

In addition, it adjusts resources for tasks to accommodate real input data.

### DIFF
--- a/10x/count/count.wdl
+++ b/10x/count/count.wdl
@@ -48,7 +48,8 @@ task setup_chunks {
   runtime {
     docker: "marcusczi/cellranger_clean:cromwell"
     memory: "2 GB"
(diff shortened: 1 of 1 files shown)


ambrosejcarr:
Testing a full run without call caching + monitoring: bab0f090-555b-4e60-9963-561d54820f1d

ambrosejcarr:
This ran through successfully, so I think this is good to merge!

PR #31, opened 2017-10-10
Move wrapper wdl prep inputs code into python package

samanehsan:
- Move code to get and parse workflow inputs from the wrapper wdls into a utils file
- Include the utils file in the `submit` package, and rename that package to `secondary_analysis`


docker/secondary-analysis-python/Dockerfile
```diff
@@ -5,3 +5,9 @@ LABEL software="Python"
 LABEL description="Python with some useful libraries"
 
 RUN pip install requests
+
+RUN mkdir /tools
+
+COPY tools /tools
```
ambrosejcarr: If you think this is mature enough to release an alpha version, then lets put this package up on pypi; you'll need to update setup.py to include tags see: [unfortunately-also-named-cromwell-tools](https://github.com/ambrosejcarr/cromwell_tools/blob/master/setup.py), but otherwise there isn't much overhead. The result will be you no longer need to copy in a local folder, and can pip install it.

docker/secondary-analysis-python/tools/file_utils/file_utils/utils.py
```diff
@@ -0,0 +1,28 @@
+import requests
+
+
+def get_file_by_uuid(file_id):
```
ambrosejcarr: I like to include `verbose` flags in the definition of function like these in case one might want to interact with stdout or silence printing for some reason later. 

How do you feel about adding that? 

Here it would just be: 
```
def get_file_by_uuid(file_id, verbose=True):
    url = 'https://dss.dev.data.humancellatlas.org/v1/files/{}?replica=gcp'.format(file_id)
    response = requests.get(url)
    if verbose:
        print('GET {0}'.format(url))
        print('{0}'.format(response.status_code))
        print('{0}'.format(response.text))
    return response.json()
```

which would have the exact same invocation, but adds a bit more reusability.

docker/secondary-analysis-python/tools/file_utils/file_utils/utils.py
```diff
@@ -0,0 +1,28 @@
...
+    url = 'https://dss.dev.data.humancellatlas.org/v1/files/{}?replica=gcp'.format(file_id)
+    print('GET {0}'.format(url))
+    response = requests.get(url)
+    print('{0}'.format(response.status_code))
+    print('{0}'.format(response.text))
+    return response.json()
+
+
+def get_manifest_files(bundle_uuid, bundle_version):
+    url = "https://dss.dev.data.humancellatlas.org/v1/bundles/" + bundle_uuid + "?version=" + bundle_version + "&replica=gcp&directurls=true"
```
ambrosejcarr: should this version `v1` be a parameter (version=`v1`) -- do we expect them to change this? 
```
def get_manifest_files(bundle_uuid, bundle_version, api_version='v1'):
    url = ("https://dss.dev.data.humancellatlas.org/{api_version}/bundles/{uuid}?version="
           "{bundle_version}&replica=gcp&directurls=true".format(
               api_version=api_version, uuid=bundle_uuid, bundle_version=bundle_version))
```

samanehsan: I'm not sure about the api versioning, but I can check in with @dshiga about that!

samanehsan: Also now that I'm looking at this again I think the dss url should also be a parameter to easily switch between staging and dev.

docker/secondary-analysis-python/Dockerfile
samanehsan: That's a great point! I wasn't sure whether to keep this as a separate package, or include it as part of #19 so I just kept it local for now

docker/secondary-analysis-python/tools/file_utils/file_utils/utils.py
ambrosejcarr: Great!

samanehsan: @dshiga mentioned the api version won't change frequently, so I included that as part of the `dss_url` parameter (e.g. "https://dss.dev.data.humancellatlas.org/v1")

samanehsan:
Done responding to your comments @ambrosejcarr! :octocat:

docker/secondary-analysis-python/tools/file_utils/file_utils/utils.py
```diff
@@ -0,0 +1,30 @@
...
+        print('{0}'.format(response.text))
+    return response.json()
+
+
+def get_manifest_files(bundle_uuid, bundle_version, dss_url, verbose=True):
+    url = '{dss_url}/bundles/{bundle_uuid}?version={bundle_version}&replica=gcp&directurls=true'.format(
+        dss_url=dss_url, bundle_uuid=bundle_uuid, bundle_version=bundle_version)
+    response = requests.get(url)
+    if verbose:
+        print('GET {0}'.format(url))
```
ambrosejcarr: print instead of '{}'.format(print)

docker/secondary-analysis-python/tools/file_utils/file_utils/utils.py
```diff
@@ -0,0 +1,30 @@
...
+        print(response.status_code)
```
rexwangcc: Just a personal preference, logging may be better than `print() ` sometimes.

docker/secondary-analysis-python/tools/file_utils/setup.py
```diff
@@ -0,0 +1,15 @@
+from setuptools import setup
+
+setup(name='file_utils',
+      version='1.0.0.dev1',
+      description='Utilities for retrieving files from the HCA-DCP storage service',
+      url='http://github.com/HumanCellAtlas/skylab',
+      author='Saman Ehsan',
+      author_email='sehsan@broadinstitute.org',
+      license='MIT',
```
rexwangcc: I'm not sure if the license needs to be consistent, but skylab git repo is under `BSD 3-clause "New" or "Revised" License`, not sure if we need to modify `MIT` to `BSD 3-clause "New" or "Revised" License` here.

samanehsan: Thanks for pointing that out @rexwangcc!

samanehsan:
Thanks for catching that @rexwangcc! I just fixed the imports 🍂

10x/hca-dcp/wrapper_10x_count.wdl
```diff
@@ -10,55 +10,27 @@ task GetInputs {
...
-    import json
-    import requests
-    import subprocess
-    import time
+    from secondary_analysis import utils
 
     # Get bundle manifest
     uuid = '${bundle_uuid}'
     version = '${bundle_version}'
+    dss_url = '${dss_url}'
```
ambrosejcarr: I can't decide if I like having these variable conversions or not. On one hand, it makes it easier if we want to change the name of the wdl variable. On the other hand, it makes it less clear in the code where the wdl variable is being used and adds lines. 

@dshiga I see your commit originally started doing these conversions. Interested in hearing why if you have time :)

docker/secondary-analysis-python/Dockerfile
```diff
@@ -12,4 +12,4 @@ RUN mkdir /tools
 
 COPY tools /tools
 
-RUN pip install -e '/tools/submit'
+RUN pip install -e '/tools/secondary_analysis'
```
ambrosejcarr: I'm wondering why we made the decision to install in develop mode instead of doing a regular `pip install setup.py`. Was it to avoid using a `MANIFEST.in` to copy over non-python files?

docker/secondary-analysis-python/tools/secondary_analysis/secondary_analysis/create_analysis_json.py
```diff
@@ -123,9 +123,9 @@ def main():
     parser.add_argument('-run_type', required=True, help='Should always be "run" for now, may be "copy-forward" in some cases in future')
     parser.add_argument('-method', required=True, help='Supposed to be method store url, for now can be url for wdl in skylab')
     parser.add_argument('-schema_version', required=True, help='The metadata schema version that this analysis.json conforms to')
-    parser.add_argument('-inputs_file', required=True, 'Path to tsv containing info about inputs')
-    parser.add_argument('-outputs_file', required=True, 'Path to json file containing info about outputs')
-    parser.add_argument('-format_map', required=True, 'JSON file providing map of file extensions to formats')
+    parser.add_argument('-inputs_file', required=True, help='Path to tsv containing info about inputs')
```
ambrosejcarr: wow did this even work before?

docker/secondary-analysis-python/tools/secondary_analysis/secondary_analysis/create_analysis_json.py
```diff
@@ -123,9 +123,9 @@ def main():
     parser.add_argument('-run_type', required=True, help='Should always be "run" for now, may be "copy-forward" in some cases in future')
```
ambrosejcarr: This doesn't matter that much, but the convention I normally see in python is `--long-argument` and `-a` for short single-letter arguments.

docker/secondary-analysis-python/tools/secondary_analysis/secondary_analysis/utils.py
```diff
@@ -0,0 +1,43 @@
+import requests
+import logging
+import time
+
+
+def get_file_by_uuid(file_id, dss_url):
```
ambrosejcarr: If possible it would be great to throw some docstrings on these with a short description + parameter types. 

It _really_ makes it easier to go back in and see what the variable types are. :)

docker/secondary-analysis-python/Dockerfile
dshiga: Do you mean instead of doing `python setup.py install`? I think we ended up doing it this way simply because I was trying to use pip to install it and `pip install -e dir` is the way the interwebs tell you to do it: https://stackoverflow.com/questions/15031694/installing-python-packages-from-local-file-system-folder-with-pip

Is there an advantage to doing it another way, like `python setup.py install` instead?

10x/hca-dcp/wrapper_10x_count.wdl
dshiga: If I understand what you're asking, I did it this way because I needed to use the uuid and version in multiple places in the Python code and didn't want to do the conversion in multiple places. So I did it once at the beginning and assigned the results to Python variables.

I see what you mean, but I prefer doing it this way because it provides a cleaner handoff from WDL to Python code, as opposed to sprinkling conversions through more of the code. An even cleaner interface would be to make the Python invokable as a script and make all WDL variables command line parameters to that script. The downside there is that it's a bit more complicated, so I don't think it's really necessary here, though I like that pattern overall.

docker/secondary-analysis-python/Dockerfile
ambrosejcarr: The biggest functional difference that I've seen is that `-e` or develop mode soft-links to the directory that you're installing from: changes to that directory will be reflected in your code. 

`python setup.py install` installs into your `site-packages directory` (and leaves behind anything not in `MANIFEST.in`. 

Here I doubt there is a, but `python setup.py install` is a "normal" way to install the package, and is a bit more strict about how the package contents is defined in `setup.py`. If we go with this way and everything works, there are slightly better guarantees of portability.

10x/hca-dcp/wrapper_10x_count.wdl
ambrosejcarr: Awesome, that's exactly what I wanted to learn. Thanks for the explanation @dshiga

docker/secondary-analysis-python/Dockerfile
dshiga: Okay, changing to `python setup.py install` sounds like a worthwhile change. @samanehsan do you think you could experiment with that and see if you can put it in this PR? If it turns out to be a lot of work we should put it in a separate ticket and PR.

samanehsan:
Thanks for the feedback @rexwangcc @ambrosejcarr @dshiga!

Issue #28, opened 2017-10-07
Refactor cost script to take server + authentication as parameters


ambrosejcarr:
`skylab/smartseq2_single_sample/run_multi_samples/scripts/calculate_cost_workflow.py` currently hardcodes the server name, username, and password. 

Failing for me (running my own `cromwell 29` server) with the following traceback: 

```
Traceback (most recent call last):
  File "/Users/carra1/projects/skylab/smartseq2_single_sample/run_multi_samples/scripts/calculate_cost_workflow.py", line 248, in <module>
    main()
  File "/Users/carra1/projects/skylab/smartseq2_single_sample/run_multi_samples/scripts/calculate_cost_workflow.py", line 246, in main
    parse_metadata_json(args.input_json_file)
  File "/Users/carra1/projects/skylab/smartseq2_single_sample/run_multi_samples/scripts/calculate_cost_workflow.py", line 215, in parse_metadata_json
    username = meta['username']
KeyError: 'username'
```

jishuxu:
Just want to clarify that the input of this script is a json file which list of subworkflow name, server name and usename, password. it is not hardcoded in this script.

ambrosejcarr:
You're right, sorry; I was imprecise. It does hardcode the server name, though. I was trying to use this on a cromwell other than ours. I'll see if I can adjust it.

ambrosejcarr:
I think I had a misunderstanding about this script: it is specifically designed to capture a subworkflow's cost in absence of the parent workflow, is that correct? 

If one wanted to, for example, run this on SS2, one would run this script once per subworkflow, is that right?

kbergin:
Working on porting tickets to zenhub, so seeing old issues. Is this still relevant?

ambrosejcarr:
Per offline conversations, this script is specific to SS2 cost analysis. I think we can close it.

PR #33, opened 2017-10-16
added input file for benchmarking runs

ambrosejcarr:
I'd like to put the inputs JSON file into the 10x count folder so it can be referenced in the wiki as the input that was run to benchmark the pipeline. 

I've also uploaded a runnable notebook to do 10x cost analysis, the requirements for the notebook, and a README file. This is mostly just as a test to see if we like this format. It would be very easy to convert this into a script instead, if that's preferable.


10x/benchmark/10x_cost_analysis.ipynb
```diff
@@ -0,0 +1,295 @@
...
...3gu4q20P4MuuXAysLukpg4otIiIi\n2m1K5+BI2gO4w/aVI06tC/y66/j2um20f+NASQskLVi4cOGAIo2IiIiSTVmCI2lF4P3Ahyby79g+\nzvZc23NnzZo1OcFFREREqyw7hT9rI2AOcKUkgPWAyyRtDdwBrN913/XqtoiIiIhxm7IeHNtX217L\n9mzbs6mGoba0/VvgDOCN9WqqbYH7bN85VbFFREREuwxymfgpwEXAppJul3TAUu7+PeBm4Cbgi8Bb\nBxVXREREtN/Ahqhs7zvG+dldtw0cPKhYIiIiYnpJJeOIiIhonSQ4ERER0TpJcCIiIqJ1kuBERERE\n6yTBiYiIiNZJghMRERGtkwQnIiIiWicJTkRERLROEpyIiIhonSQ4ERER0TpJcCIiIqJ1kuBERERE\n6yTBiYiIiNZJghMRERGtkwQnIiIiWicJTkRERLROEpyIiIhonYElOJJOkHS3pGu62v5D0g2SrpL0\nTUmrd507QtJNkm6U9PJBxRURERHtN8genJOAXUa0nQ08y/bmwP8DjgCQtBmwD/DM+jFfkLTMAGOL\niIiIFhtYgmN7PvCHEW1n2V5UH14MrFff3gP4qu0Hbd8C3ARsPajYIiIiot2anIPzZuD79e11gV93\nnbu9bnscSQdKWiBpwcKFCwccYkRERJSokQRH0geARcDJ432s7eNsz7U9d9asWZMfXERERBRv2an+\ngZL+AXglsJNt1813AOt33W29ui0iIiJi3Ka0B0fSLsB7gd1t/7nr1BnAPpKeIGkOsDFwyVTGFhER\nEe0xsB4cSacALwbWlHQ7cCTVqqknAGdLArjY9j/bvlbSqcB1VENXB9t+eFCxRURERLsNLMGxve8o\nzccv5f4fBT46qHgiIiJi+kgl44iIiGidJDgRERHROklwIiIionWS4ERERETrJMGJiIiI1kmCExER\nEa2TBCciIiJaJwlOREREtE4SnIiIiGidJDgRERHROklwIiIionWS4ERERETrJMGJiIiI1kmCExER\nEa2TBCciIiJaJwlOREREtM7AEhxJJ0i6W9I1XW1rSDpb0i/q7zPrdkn6rKSbJF0lactBxRURERHt\nN8genJOAXUa0HQ6ca3tj4Nz6GGBXYOP660Dg2AHGFRERES03sATH9nzgDyOa9wDm1bfnAXt2tX/Z\n
```
rexwangcc: I feel this notebook is super useful, especially the $244.01 here along with the bar plots.

10x/benchmark/10x_cost_analysis.ipynb
```diff
@@ -0,0 +1,295 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "%%bash\n",
+    "# download cost script\n",
+    "curl -o ${TMPDIR}get_cost.py https://raw.githubusercontent.com/ambrosejcarr/cromwell-manager/master/src/accessories/calculate_workflow_cost.py 2>&1 > /dev/null\n",
+    "\n",
+    "# run the cost script on the metadata\n",
+    "python2.7 ${TMPDIR}/get_cost.py -m ${METADATA} > ./cost_data.txt"
```
rexwangcc: I'm not sure if we need to convert python2.7 to python3 for get_cost.py later in the future?

rexwangcc: Sorry, it seems this get_cost.py is downloaded from another repo of yours. So it won't be an issue as long as it works.

10x/benchmark/10x_cost_analysis.ipynb
```diff
@@ -0,0 +1,295 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "# to run interactively, uncomment the line below\n",
+    "# os.environ['METADATA] = 'path_to_metadata.json'"
```
rexwangcc: Is this a typo? I assume it would be `os.environ['METADATA'] = './metadata.json'`

10x/benchmark/10x_cost_analysis.ipynb
ambrosejcarr: Correct; it's a command line utility being called by your local environment's python2.7. I put that script into my repo because I was getting authentication errors pulling it from the original dsde repo.

10x/benchmark/10x_cost_analysis.ipynb
ambrosejcarr: Yes! great catch. I'll fix this.

10x/benchmark/10x_cost_analysis.ipynb
dshiga: We already have a copy of the cost script in skylab here: 
https://github.com/HumanCellAtlas/skylab/blob/master/smartseq2_single_sample/run_multi_samples/scripts/calculate_cost_workflow.py

Could you please use that one for now? We have a ticket to put that script in cromwell-tools, so later we can reference it from there.

ambrosejcarr:
Dave, thanks!

This is the type of conversation I wanted to have about this method of code sharing. I think having it run as a script is fine as well. In this case, my first attempt would be to generate a report with a brief explanation of the two graphs, as well as the resulting total cost and time at the bottom. 

The reason I like notebooks is two-fold:

1. It makes documentation a first-class citizen. Markdown cells in a notebook tend to get read when executed, whereas documentation in a script tends to be read when things get break. I think of notebooks as places to demonstrate less mature code that you don't necessarily see as "accepted" -- maybe someone wants to criticize the way I've coded something. Long term, it makes MUCH more sense to have re-runnable code in scripts than in notebooks. 
2. Images. In my opinion it's much nicer to have an image with a caption plus the code used to generate it than just get the output figure at the end. 

Regarding the cost script checked in to skylab, I made an issue about this #28 -- @jishuxu converted it to python3, but she also hard-coded a few variables like server, password, and uername and I wasn't clear on how to use it or configure it (it breaks for me by default). As a result, I'm using the vanilla DSDE cost script (also checked in, since the pipelines repo has restrictive permissions)

Regarding code usability, I do think the way of estimating cost and time from the metadata is potentially reusable. It might make sense to pull that into the repo I was demoing this morning and extract it from the notebook. The rest is not very portable -- it fixes problems with the split/merge/join being done explicitly in count.wdl instead of using cromwell/wdl syntax. 

Thank you so much for your thoughtful comments!

ambrosejcarr:
Leaving some notes from conversation with @dshiga here: 
- interest in extracting any reusable code from this notebook to the calculate_cost script (such as the run timing information)
- potential to revisit this later to transform into a script that generates a report
- when skylab cost script is generalized or moved to cromwell-tools, edit this notebook to point to that script instead.

PR #36, opened 2017-10-21
uBAMs work as input to STAR

ambrosejcarr:
Added several files to facilitate testing of uBAM inputs to STAR. Successful tests are displayed in test.wdl. Star succesfully: 

1. Aligns uBAMS
2. Propagates tags

To test this, I generated some task wdls to make references from the new version of star (not backwards compatible), align fastq reads using the same star version for comparison, tag bams from fastq files in order to check tag propagation, and align uBAM reads with STAR. This also required some new dockerfiles. These files are listed categorically below. 

I believe the task wdls are reusable. The analysis wdl is not, so I have checked it into an analysis directory under optimus. In addition to a normal code review, I'm specifically interested in input on file organization and whether each should all be checked in. 

Dockerfiles: 

1. python3-scientific (updated scsequtil version)
2. kallisto
3. star (from master to test ubams)

Task wdls:

1. Attach10xBarcodes.wdl
2. FastqToUBam.wdl
3. StarAlignBamSingleEnd.wdl
4. StarMkref.wdl
5. StarAlignBamSingleEnd.wdl

Workflow wdls:

1. test_ubam_alignment.wdl

Testing notebooks:

1. optimus/analysis/test_star/test.ipynb


optimus/wdl_tasks/StarMkref.wdl
```diff
@@ -0,0 +1,28 @@
+
+task StarMkref {
```
dshiga: There is overlap here between this and https://github.com/HumanCellAtlas/skylab/blob/master/smartseq2_single_sample/build_references/build_star_reference/BuildStarReferenceBundle.wdl

There might be a good reason to keep them separate, but worth thinking about whether we can consolidate.

optimus/wdl_tasks/FastqToUBam.wdl
```diff
@@ -0,0 +1,28 @@
+
+task FastqToUBam {
+  File gz_fastq_file  # input fastq file
+  String sample_name  # name of sample matching this file, inserted into read group header
+
+  # estimate that bam is approximately equal in size to fastq, add 20% buffer
+  Int estimated_disk_required = ceil(size(gz_fastq_file, "G") * 2.2)
```
dshiga: In general I think we should be very generous with resource allocation, then benchmark and optimize where needed. For production pipelines, not having enough wiggle room will cause lots of headaches with failed workflows, even if it only happens for 1 in 1000 bundles. If this is a very well behaved, predictable task in terms of disk requirements then 20% is probably fine. If we're less sure, then I would err more on the side of being too generous and give, say 2x as much as we typically need.

optimus/wdl_tasks/Attach10xBarcodes.wdl
```diff
@@ -0,0 +1,31 @@
+
+task Attach10xBarcodes {
+  File r1
+  File i7
+  File u2
```
dshiga: I am liking the comments you put on inputs for other tasks. Could you do that here as well?

optimus/wdl_tasks/StarAlignFastqSingleEnd.wdl
```diff
@@ -0,0 +1,39 @@
...
+  Int estimated_disk_required =  ceil(input_size * 2.2 + reference_size)
+
+  command {
+    # prepare reference
+    mkdir genome_reference && tar -xzf ${gztar_star_reference} -C genome_reference --strip-components 1
+    rm ${gztar_star_reference}
+
+    STAR \
+      --runMode alignReads \
+      --runThreadN 16 \
```
mckinsel: If you replace 16 with `$(nproc)` you won't have to worry about remembering to change this if you change the instance type.

optimus/wdl_tasks/Attach10xBarcodes.wdl
```diff
@@ -0,0 +1,31 @@
...
+
+  # estimate disk requirements
+  Float bam_size = size(u2, "G")
+  Float fastq_size = size(r1, "G") + size(i7, "G")
+  Int estimated_required_disk = ceil(fastq_size + bam_size * 2.2)
+  # other parameters
```
dshiga: There are no other parameters right now, so I think this comment can be deleted.

optimus/wdl_tasks/StarAlignFastqSingleEnd.wdl
```diff
@@ -0,0 +1,39 @@
...
+
+  # estimate input requirements for star using UBams:
+  # multiply input size by 2.2 to account for output bam file + 20% overhead, add size of reference.
+  Int input_size = ceil(size(fastq_input, "G"))
+  Int reference_size = ceil(size(gztar_star_reference, "G"))
+  Int estimated_disk_required =  ceil(input_size * 2.2 + reference_size)
+
+  command {
+    # prepare reference
+    mkdir genome_reference && tar -xzf ${gztar_star_reference} -C genome_reference --strip-components 1
```
mckinsel: If you drop the `z` flag it should work with both compressed and uncompressed tarballs.

optimus/wdl_tasks/StarMkref.wdl
ambrosejcarr: I guess the big difference is that I needed to run a different version of star. We could pass the name of the docker image as a parameter and then I expect the two should be mergeable. 

it brings up a larger organizational question about how and where we put the tasks: perhaps we need a "general" location, as well as folders for 3' and full-length specific tasks?

optimus/wdl_tasks/FastqToUBam.wdl
ambrosejcarr: Message received: in the general case, err on the side of running the task without problems instead of over-optimizing to save money, but inducing operational problems. 

In this case the majority of the information in the two file types is text data (read name, sequence, and quality) which is present in both file types, and both types us gzip compression, so the types are quite consistently similar in size. I could boost this up a little bit, but 2x is definitely not required for this particular task.

optimus/wdl_tasks/Attach10xBarcodes.wdl
ambrosejcarr: 👍

optimus/wdl_tasks/StarAlignFastqSingleEnd.wdl
ambrosejcarr: Great idea. Will change.

optimus/wdl_tasks/Attach10xBarcodes.wdl
ambrosejcarr: 👍

optimus/wdl_tasks/StarAlignFastqSingleEnd.wdl
ambrosejcarr: I didn't know that. Thank you for the suggestion.

... in what circumstance do you need to pass `-j` or `-z` to `tar -xf`?

ambrosejcarr:
Those are good point. I couldn't diff them because the tags are different so I was sorting and comparing, but didn't put that code into the notebook. I'll add it in, and will change the directory name to tasks.

optimus/wdl_tasks/StarAlignFastqSingleEnd.wdl
mckinsel: I think if your version of tar is more than 13 years old
https://www.gnu.org/software/tar/

ambrosejcarr:
@dshiga I moved/renamed a bunch of stuff to fit the new organization we discussed. 

@mckinsel any further comments? Happy to have you review every PR where I use bash. I feel like you're slowly teaching me the language. ;)

pipelines/tasks/StarMkref.wdl
```diff
@@ -0,0 +1,31 @@
+
+task StarMkref {
+  File fasta_file  # genome annotation
+  File annotation_file  # gtf annotation file
+
+  # optionally, use a different version of star
+  String star_docker_image = "ambrosejcarr/star-ubam:v1.1.0"
```
mckinsel: Would it make sense to have the STAR version in the tag somewhere?

pipelines/tasks/StarMkref.wdl
```diff
@@ -0,0 +1,31 @@
...
+  
+  command {
+    mkdir genome
+    STAR \
+      --runMode genomeGenerate \
+      --runThreadN $(nproc) \
+      --genomeDir genome \
+      --genomeFastaFiles ${fasta_file} \
+      --sjdbGTFfile ${annotation_file} \
```
mckinsel: There are a few of these escaped final newlines in this PR. It's not a big deal, but if somebody tries to add a command after this in the future it's going to break.

pipelines/tasks/Attach10xBarcodes.wdl
```diff
@@ -0,0 +1,30 @@
...
+  File u2  # reverse bam file; contains alignable genomic information
+
+  # estimate disk requirements
+  Float bam_size = size(u2, "G")
+  Float fastq_size = size(r1, "G") + size(i7, "G")
+  Int estimated_required_disk = ceil(fastq_size + bam_size * 2.2)
+
+  command {
+    Attach10xBarcodes \
+      --r1 ${r1} \
```
mckinsel: You generally want to double-quote all of these things. There's some discussion [here](http://mywiki.wooledge.org/Quotes).

docker/python3-scientific/Dockerfile
```diff
@@ -4,7 +4,7 @@ MAINTAINER Ambrose Carr <mail@ambrosejcarr.com>
 LABEL software="python 3.6.2"
 LABEL description="python 3.6.2 with pysam and scsequtil"
 
-RUN pip3 install scsequtil==0.0.5 \
+RUN pip3 install scsequtil==0.0.6 \
```
mckinsel: You might want to define versions for the subsequent packages too. Changes in versions of numpy or scipy can create lots of problems.

mckinsel:
@ambrosejcarr, I added a couple more minor comments. Nothing critical. One general question: does it matter that the input BAM has no stored alignments?

pipelines/tasks/StarMkref.wdl
```diff
@@ -0,0 +1,31 @@
...
+  
+  runtime {
+    docker: "${star_docker_image}"
+    cpu: 16
+    memory: "30 GB"
+    disks: "local-disk 100 SSD"
+  }
+  
+  output {
+    File genome = "genome.tar.gz"
```
jishuxu: so you are going to tar.gz reference then? I remember the gz file will take extra time compare to tar file.

ambrosejcarr: We should test how long these each take. In my experience, while .gz takes more time to decompress, the reduced size makes them transfer faster and the transfer speeds I've been seeing to GCE instances is _slow_.

jishuxu: If I remember correctly, the size difference bwt tar.gz and gz of reference bundle is not significant and but without  unzipping process, it will save about 10~15 min on SS2. But SS2 pipeline is  single-sample per run fashion so gz or un-gz is more significant than 10X I guess.

ambrosejcarr: I'll explicitly test. If there is a negative impact or mild positive impact for mine I'll remove the gzip since it sounds like it makes things slower for SS2.

mckinsel: You can often separate and parallelize the compression to speed things up. Something like `tar cf - genome/* | pigz > genome.tar.gz`

mckinsel: Also, is there any ability to do streaming with WDL/cromwell? Ideally you would be able to do something like
```
stream-input ${input_tar} | gunzip | tar xf - -C genome
```
and then the unzipping and the file transfer can happen simultaneously. But I don't think there's anything like `stream-input` in WDL that would work without abusing types.

ambrosejcarr: I don't think this exists. I really wish it would, though. It would make tasks so much faster...

ambrosejcarr: pigz -d is a bit slower (~10%) than the download speed I'm getting on the instances. I've removed the gzipping. Thanks for the comment!

ambrosejcarr:
I think I've addressed all the comments. Thanks for the bash help @mckinsel 

Regarding your streaming input question Marcus, it seems there are people working on that. There doesn't appear to be a problem with passing in strings and streaming in data with gsutil cp.

PR #38, opened 2017-10-27
Ss2 check reference

jishuxu:
in this PR,
1) build simple workflow to align example dataset and quantify gene counts. 
2) run analysis of STAR logs and gene counts data to understand the impact from different reference bundles.


smartseq2_single_sample/check_reference/analysis/reference_check.ipynb
```diff
@@ -0,0 +1,561 @@
+{
```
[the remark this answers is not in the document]
ambrosejcarr: For (2), finalize gene biotypes included in reference:

I think the best thing to do here would be to lean on the experts at gencode and refseq. They seemed willing to help and are probably much better versed at identifying which biotypes we should see and which we shouldn't. 

A few things piqued my interest during the call: 
1. some small RNAs are detectable because they're carried within larger lincRNA. It sounds like at least refseq does not have any annotation to partition these from those that are directly transcribed as short hairpins and therefore are too small to detect. We should not deplete these. 
2. Transcribed pseudogenes are a biotype that I wasn't aware of. These could be included but would result in multi-mapping of those genes. We should investigate this further.

smartseq2_single_sample/check_reference/buildRef.wdl
```diff
@@ -0,0 +1,20 @@
+import "BuildStarReferenceBundle.wdl" as buildStarRef
+
+workflow buildStarRefs {
+  File ref_metadata_file
+  Array[Object] metadata = read_objects(ref_metadata_file)
+  
+  scatter (record in metadata){
+    String name_ref = record.name1
+    String name2 = record.name2
```
ambrosejcarr: Is this not used? Also It would help to have a comment here or a more informative name on this and also name_ref.

smartseq2_single_sample/check_reference/buildRef.wdl
```diff
@@ -0,0 +1,20 @@
...
+    File fasta = record.fasta
+    File gtf = record.gtf
+    
+    call buildStarRef.BuildStarReference as StarRef {
+      input:
+        ref_fasta = fasta,
+        gtf_file = gtf,
+        ref_name = name_ref
```
ambrosejcarr: I feel like inverting the name in this way (`ref_name <-> name_ref`) makes downstream maintenance a little trickier. Maybe the SWE who reviews this can comment.

smartseq2_single_sample/check_reference/reference_check_wf.wdl
```diff
@@ -0,0 +1,43 @@
+import "star.wdl" as RunStar
+import "featurecounts.wdl" as fc
+
+workflow CheckReference {
+
+  File input_metadata
+  Array[Object] metadata = read_objects(input_metadata)
```
ambrosejcarr: Another input object that I would unpack.

smartseq2_single_sample/check_reference/buildRef.wdl
```diff
@@ -0,0 +1,20 @@
+import "BuildStarReferenceBundle.wdl" as buildStarRef
+
+workflow buildStarRefs {
+  File ref_metadata_file
+  Array[Object] metadata = read_objects(ref_metadata_file)
```
ambrosejcarr: Unless it makes this more modular or easier to reuse, I find that this kind of input metadata abstraction (Array[Object]) makes the code more opaque. I would much rather have explicit inputs to the function in terms of individual arrays so I can see the inputs being scattered over.

smartseq2_single_sample/check_reference/reference_check_wf.wdl
```diff
@@ -0,0 +1,43 @@
...
+  String sra_dir = "gs://broad-dsde-mint-dev-teststorage/patel_ap/"
```
ambrosejcarr: I wouldn't hard-code this. It would make this really hard to reuse on different data. An alternative would be to pass the full links in 13-20 as inputs instead of constructing them within the wdl.

smartseq2_single_sample/check_reference/reference_check_wf.wdl
```diff
@@ -0,0 +1,43 @@
...
+    call fc.FeatureCountsMultiMapping as multcount {
+      input:
+        aligned_bam=Star.output_bam,
+        gtf = record.gtf,
+        fc_out = record.sraID+"_"+record.length+"_"+record.ref_name
+    }
+ }
+
+ output {
+   Star.*
```
ambrosejcarr: I think if using .* I would maybe comment what outputs you're using here; most people will go to the workflow wdl first, and since we have our tasks in individual files, they would then have to open up each individual file to figure out what is output here.

smartseq2_single_sample/check_reference/tasks/featurecounts.wdl
```diff
@@ -0,0 +1,46 @@
...
+  String fc_out
+  
+  command {
+    featureCounts -T 4  -s 0 -t exon -g gene_id -p -B -C -a ${gtf} -o "${fc_out}.gene.unq.counts.txt" ${aligned_bam}
+    featureCounts -T 4  -s 0 -t exon -g transcript_id -p -B -C -a ${gtf} -o "${fc_out}.transcript.unq.counts.txt" ${aligned_bam}
+    featureCounts -T 4  -s 0 -t exon -g exon_id -p -B -C -a ${gtf} -o "${fc_out}.exon.unq.counts.txt" ${aligned_bam}
+  }
+  runtime {
+    docker:"humancellatlas/star_dev:v1"
+    memory: "4 GB"
```
ambrosejcarr: This cpu count / runtime isn't efficient. 4 cpu should net you 16gb ram. Does this need more than 1 cpu? 3.75 gb is pretty close to 4. :-)

Can also apply to the below task.

smartseq2_single_sample/check_reference/tasks/picard.wdl
```diff
@@ -0,0 +1,56 @@
...
+      ASSUME_SORTED=true
+
+      tar -cvf "${output_file}.tar" ${output_file}.*.txt
+  }
+  output {
+    File alignment_metrics = "${output_filename}.tar"
+  }
+  runtime {
+    docker:"broadinstitute/genomes-in-the-cloud:2.3.1-1504795437"
+    memory:"4 GB"
```
ambrosejcarr: saves money to use 3.75gb if you're able to. Otherwise you might as well use 7.5gb. There are a few more cases across the PR where I would look into this.

smartseq2_single_sample/check_reference/tasks/featurecounts.wdl
```diff
@@ -0,0 +1,46 @@
+task FeatureCountsUniqueMapping {
+  File aligned_bam
+  File gtf
+  String fc_out
+  
+  command {
+    featureCounts -T 4  -s 0 -t exon -g gene_id -p -B -C -a ${gtf} -o "${fc_out}.gene.unq.counts.txt" ${aligned_bam}
```
ambrosejcarr: Marcus tells me it is generally good practice to double quote all variable expansions. His evidence seemed compelling: http://mywiki.wooledge.org/Quotes

smartseq2_single_sample/check_reference/tasks/star.wdl
```diff
@@ -0,0 +1,50 @@
...
+  File input_fastq_read2
+  File gtf
+  File star_genome
+  String sample_tag
+  String pu_tag
+  String id_tag
+  String lib_tag
+
+
+  command {
```
ambrosejcarr: In some tasks you leave spaces between command output and runtime and in others you don't. 

... I like the spaces. :-)

smartseq2_single_sample/check_reference/buildRef.wdl
jishuxu: it is why I used object in wdl. it definitely made scatter easier.

smartseq2_single_sample/check_reference/tasks/featurecounts.wdl
jishuxu: it could be faster but probably not cheaper

smartseq2_single_sample/check_reference/reference_check_wf.json
```diff
@@ -0,0 +1,3 @@
+{
```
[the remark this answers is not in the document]
jishuxu: included

smartseq2_single_sample/check_reference/tasks/star.wdl
jishuxu: I don't mind to add spaces/lines

smartseq2_single_sample/check_reference/analysis/reference_check.ipynb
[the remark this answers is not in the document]
rexwangcc: For AC 0, when I work with Jupyter notebooks, I have two ways to check typos:
1. Just copy that whole markdown text you want to check manually into Grammarly(Free) version is enough, and copy back after fixing the typos.
2. There is a great spellchecker plugin help you check typos, install it by below comments:
```bash
pip install jupyter_contrib_nbextensions # install jupyter_contrib_nbextensions
jupyter contrib nbextension install --user # copy required javascript files
jupyter nbextension enable spellchecker/main # enable the spellcheck plugin
```
For more information, you could check [here](http://jupyter-contrib-nbextensions.readthedocs.io/en/latest/install.html#install-javascript-and-css-files)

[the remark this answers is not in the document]
ambrosejcarr: Ah cool; I already have nbextensions installed. That's very easy! Thanks @rexwangcc

smartseq2_single_sample/check_reference/tasks/featurecounts.wdl
```diff
@@ -0,0 +1,47 @@
...
+  File gtf
+  String fc_out
+  
+  command {
+    featureCounts -T 4  -s 0 -t exon -g gene_id -p -B -C -a "${gtf}" -o "${fc_out}.gene.unq.counts.txt" "${aligned_bam}"
+    featureCounts -T 4  -s 0 -t exon -g transcript_id -p -B -C -a "${gtf}" -o "${fc_out}.transcript.unq.counts.txt" "${aligned_bam}"
+    featureCounts -T 4  -s 0 -t exon -g exon_id -p -B -C -a "${gtf}" -o "${fc_out}.exon.unq.counts.txt" "${aligned_bam}"
+  }
+  runtime {
+    docker:"humancellatlas/star_dev:v1"
```
rexwangcc: @jishuxu This should also be changed to "humancellatlas/star_dev:2.5.3a"

jishuxu: Just changed

smartseq2_single_sample/check_reference/reference_check_wf.json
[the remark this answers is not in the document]
ambrosejcarr: I'm aware it's in the bucket, but I more meant that having to go to a bucket to see what the inputs to your wdl were was onerous, particularly because they weren't typed in the file, but rather abstracted by the Array[Object] type.

smartseq2_single_sample/tasks/featurecounts.wdl
```diff
@@ -0,0 +1,47 @@
...
+  File aligned_bam
+  File gtf
+  String fc_out
+  
+  command {
+    featureCounts -T 4  -s 0 -t exon -g gene_id -p -B -C -a "${gtf}" -o "${fc_out}.gene.unq.counts.txt" "${aligned_bam}"
+    featureCounts -T 4  -s 0 -t exon -g transcript_id -p -B -C -a "${gtf}" -o "${fc_out}.transcript.unq.counts.txt" "${aligned_bam}"
+    featureCounts -T 4  -s 0 -t exon -g exon_id -p -B -C -a "${gtf}" -o "${fc_out}.exon.unq.counts.txt" "${aligned_bam}"
+  }
+  runtime {
```
ambrosejcarr: Some tasks have the order of outputs and runtime inverted. I suppose there is no convention for this but we should try to be consistent. I've normally put output last.

jishuxu: will change it

ambrosejcarr:
General comment about this: I *think* our last discussion suggested that the analysis files should be in `/analysis/check_reference/<... your choice of directory structure ...>`

You should check with @rexwangcc and @dshiga so that this PR can commit files in the right forward-compatible locations.

ambrosejcarr:
Please rename the analysis notebooks to make clear that notebook 1 deals with alignment and notebook 2 deals with quantification, instead of them having the same name + "I" and "II".

ambrosejcarr:
It would be nice to have a bit of interpretation on the alignment notebook. From the plots I learn that Gencode + Refseq are more or less the same (on this scale) and ensembl is a big outlier for: reads mapped to too many loci, reads mapped to multiple loci, and reads uniquely mapped. Are there any metadata from the table that help explain why ensembl is acting like an outlier in these tests? Could we do such a regression analysis? Is there anything we can learn here?

ambrosejcarr:
For the quantification notebook, I think the drop-out plots are the correct approach to investigate what's going on 👍 . I think there are a few more interesting things we could learn, and that the AWG would be interested in: 

1. What is the R2 value? That will tell us how important this effect is. If the R2 is .99, then they can basically just rubber-stamp your analysis. If it's .9 or .85 then they need to **really think** about this. 
2. Scatter plots can be really misleading when plotted with full opacity, as there could be thousands on that straight line. I would re-plot with opacity = 10% or something around that value (you can play)
3. I found it really informative to do some functional enrichments on the drop-out genes. Are they enriched for certain biotypes? certain gene ontology categories? You can use simple hypergeometric analysis on the drop-out sets or you can do GSEA on the difference in expression between the two annotations.

ambrosejcarr:
That's all I've got. This is looking more refined. Nice job, Jishu. :)

smartseq2_single_sample/tasks/picard.wdl
```diff
@@ -0,0 +1,56 @@
...
+  String output_filename
+
+  command {
+    java -Xmx4g -jar picard.jar  MarkDuplicates \
+       VALIDATION_STRINGENCY=SILENT  \
+       INPUT=${aligned_bam} \
+       OUTPUT="${output_filename}.MarkDuplicated.bam" \
+       ASSUME_SORTED=true \
+       METRICS_FILE="${output_filename}.duplicate_metrics.txt" \
+       REMOVE_DUPLICATES=false
```
samanehsan: Will `ASSUME_SORTED` and `REMOVE_DUPLICATES` ever need to have different values? If so, you might want to use variables here instead of writing in a value

smartseq2_single_sample/tasks/picard.wdl
```diff
@@ -0,0 +1,56 @@
...
+      METRIC_ACCUMULATION_LEVEL=ALL_READS \
+      INPUT=${aligned_bam} \
+      OUTPUT="${output_filename}" \
+      FILE_EXTENSION = ".txt" \
+      PROGRAM = CollectAlignmentSummaryMetrics \
+      PROGRAM = CollectInsertSizeMetrics \
+      PROGRAM = CollectGcBiasMetrics \
+      PROGRAM = RnaSeqMetrics \
+      REFERENCE_SEQUENCE=${ref_genome_fasta} \
+      ASSUME_SORTED=true
```
samanehsan: Will this always be true? If not this should be a variable

smartseq2_single_sample/tasks/picard.wdl
jishuxu: Yes, it is fixed values for us.

smartseq2_single_sample/tasks/picard.wdl
jishuxu: For us, it is always be true.

smartseq2_single_sample/check_reference/buildRef.wdl
samanehsan: Would a comment about the format of the metadata object help clarify that @ambrosejcarr?

ambrosejcarr: Yea, I think it would. Mostly I want to be able to look at a workflow wdl and see:
1. what do I need to provide this file?
2. what do I get as output.

PR #40, opened 2017-11-07
Draft: Add wdl deploy script

dshiga:
This adds a script that will copy wdls to a directory structure in a bucket that incorporates version numbers.

This will help us automate deployment, reduce manual copy errors, and avoid overwriting previous versions of wdl files.

Infrastructure / infra_version refers to the pipeline-tools repo where we will eventually put wrapper wdls.


dshiga:
We are changing how the listener gets WDLs, so there will no longer be a need to deploy them.

PR #42, opened 2017-11-08
added examination of barcode fragmentation

ambrosejcarr:
I've uploaded a notebook suggesting that there is some kind of error happening in 10x that's not explained by simple error models. It seems likely that barcode truncation could be this error, but in-depth follow-up would be required to confirm.


jishuxu:
Good to go! Approved

ambrosejcarr:
Jishu had a great comment that we should make sure that the test I ran wasn't dataset specific. In addition to the first dataset, I also ran the notebook on our pbmc8k gold standard data; the results look more or less the same.

PR #44, opened 2017-11-13
Cw pass docker image as a parameter to submit wdl 296

rexwangcc:
The get_metadata task in the `submit.wdl` uses a different docker image for the staging vs. dev deployments. This PR passes in the docker image as a parameter to this task so the same version of the submit.wdl can be used when running workflows on both staging and dev.

Also modified wrapper wdls and sample inputs to adapt to the new parameter.

### DIFF
--- a/10x/hca-dcp/wrapper_10x_count.wdl
+++ b/10x/hca-dcp/wrapper_10x_count.wdl
@@ -215,7 +215,7 @@ workflow Wrapper10xCount {
   Int timeout_seconds
 
   # Set Runtime Environment
-  String runtime_enviroånment
+  String runtime_environment
 
   call GetInputs {
     input:
--- a/10x/hca-dcp/wrapper_10x_count.wdl
+++ b/10x/hca-dcp/wrapper_10x_count.wdl
@@ -63,7 +63,7 @@ task GetInputs {
     CODE
   >>>
   runtime {
-    docker: "humancellatlas/secondary-analysis-python:0.1.2"
+    docker: "humancellatlas/secondary-analysis-python:0.1.4"
   }
   output {
     Array[File] r1 = read_lines("r1.tsv")
@@ -97,7 +97,7 @@ task rename_files {
     CODE
   >>>
   runtime {
-    docker: "humancellatlas/secondary-analysis-python:0.1.2"
+    docker: "humancellatlas/secondary-analysis-python:0.1.4"
   }
   output {
     File r1_new = "${r1_name}"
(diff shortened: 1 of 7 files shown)


hca-dcp/submit.wdl
```diff
@@ -23,7 +24,9 @@ task get_metadata {
       "$(cut -f3 $creds)/api/workflows/v1/$(cat workflow_id.txt)/metadata" > metadata.json
   >>>
   runtime {
-    docker: "gcr.io/broad-dsde-mint-dev/cromwell-metadata:0.1.0"
+    docker: if runtime_environment == "dev" then "gcr.io/broad-dsde-mint-dev/cromwell-metadata:0.1.0"
+            else if runtime_environment == "staging" then "gcr.io/broad-dsde-mint-staging/cromwell-metadata:0.1.0"
+            else "We cannot recognize the environment: " + runtime_environment
```
rexwangcc: @dshiga Recommended me that I could simplify this if-else statement to a better style such as: `"gcr.io/broad-dsde-mint-" + runtime_environment + "/cromwell-metadata:0.1.0"`

10x/hca-dcp/wrapper_10x_count.wdl
```diff
@@ -214,6 +214,9 @@ workflow Wrapper10xCount {
   Int retry_seconds
   Int timeout_seconds
 
+  # Set Runtime Environment
```
dshiga: Would be helpful to be more specific about what this is used for. Might not be obvious to someone running the wrapper that this is needed inside submit.wdl to get Cromwell metadata.

hca-dcp/submit.wdl
dshiga: Yes, the benefit is that it's much simpler, so less chance of introducing a bug, and it's easier for other developers to read and understand. If someone passes an environment other than dev or staging, it will break this task and should be pretty straightforward to debug.

hca-dcp/submit.wdl
```diff
@@ -23,7 +24,9 @@ task get_metadata {
       "$(cut -f3 $creds)/api/workflows/v1/$(cat workflow_id.txt)/metadata" > metadata.json
```
samanehsan: We can remove the api url out of the credentials file in the docker container and change this to `"https://cromwell.mint-${runtime_environment}.broadinstitute.org/api/workflows/v1/$(cat workflow_id.txt)/metadata"`. That way we can use the same docker container for both staging and dev since the url is the only difference between the two!

rexwangcc: This makes life much easier, we would then only need to maintain a smaller set of docker images! Will do it! Thanks

rexwangcc:
Tested changes for SS2 and 10x, fixed a typo at last. Going to merge it.

PR #46, opened 2017-11-15
Move secondary-analysis-python code to pipeline-tools

dshiga:
As part of repo reorganization, I'd like to move the code currently in docker/secondary-analysis-python to the new pipeline-tools repo. There is a PR to add the code to pipeline-tools here: https://github.com/HumanCellAtlas/pipeline-tools/pull/1

This PR removes that same code from this repo and updates the wdls to point to the new image. It also removes the mock smartseq2 pipeline, which we no longer need. We used it a while back to test Lira and demo things, but the real smartseq2 pipeline runs fast enough on demo data to serve the same purpose.


10x/hca-dcp/wrapper_10x_count.wdl
```diff
@@ -183,7 +183,7 @@ task inputs_for_submit {
     CODE
   >>>
   runtime {
-    docker: "humancellatlas/secondary-analysis-python:0.1.4"
+    docker: "humancellatlas/pipeline-tools:0.1.4"
```
rexwangcc: I'd like to also re-tag secondary-analysis-python from 0.1.0 to 0.1.3 images to pipeline-tools and push them to the new docker repo, and remove secondary-analysis-python docker repo then we would have a continuous version changes.

PR #49, opened 2017-11-22
Fix issues with picard docker.

rexwangcc:
For this PR, I would do two tasks:
1. Fix the `unable to access picard.jar file` issue with the dockerfile. This Dockerfile will be built and pushed to humancellatlas/picard:2.10.10 on the Docker Hub
2. Add R-base, ggplot2 and other Picard dependencies to the Dockerfile to fix the plotting function of Picard in a separate commit, and then build and push this image to humancellatlas/picard:2.10.10-xxx on the Docker Hub, `xxx` here will be the hash code of this commit.


docker/picard/Dockerfile
```diff
@@ -7,4 +7,5 @@ LABEL maintainer="Jishu Xu <jishuxu@broadinstitute.org>" \
   website="http://broadinstitute.github.io/picard"
```
ambrosejcarr: is it possible to add a quick comment to the file about how picard should be invoked?

rexwangcc: Great point! Yeah, I will add this comment, it would be really helpful to users.

rexwangcc:
Now on the Docker Hub, we have two available Picard images now;
1. humancellatlas/picard:2.10.10
2. humancellatlas/picard:2.10.10-01da379
The first one is pure picard, and the second one is with `R-base`, `reshape`, `gplots` and `ggplot2`

docker/picard/Dockerfile
jishuxu: Since the workdir is '/usr/picard' , does it mean we just need to "java -jar picard.jar" in wdl, right?

jishuxu:
overall  it is good. approved

docker/picard/Dockerfile
```diff
@@ -7,4 +7,19 @@ LABEL maintainer="Jishu Xu <jishuxu@broadinstitute.org>" \
...
 
 ENV version 2.10.10
```
jishuxu: I actually would like to move this "ENV" just before the installation of Picard.

docker/picard/Dockerfile
rexwangcc: Unfortunately, we still cannot use `java -jar picard.jar` directly. This command works for local testes, however, when we use this style for Cromwell runtime environment, it will fail. I'm guessing this may because the mechanism of Cromwell's runtime is not working as usually we do in local environments.

Testing report FYI:
1. [Works] `WORKDIR /usr/picard` in Dockerfile and `java -Xmx2g -jar /usr/picard/picard.jar FastqToSam`  [Example Workflow](https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1/d236ecbe-5b44-465a-bf5d-395a980a9fdc/timing)

2. [Not Work] `WORKDIR /usr/picard` in Dockerfile with `ENV PATH="/usr/picard:${PATH}"`and `java -Xmx2g -jar picard.jar FastqToSam` [Example Workflow](https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1/8da8578e-2eae-4ecf-897c-0fc82df3382c/timing)
3. [Not Work]  `WORKDIR /usr/picard` in Dockerfile and `java -Xmx2g -jar picard.jar FastqToSam` [Example Workflow](https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1/6b50edf6-0fad-44d5-b5b3-21fa778ca41c/timing)
4. [Not Work] `WORKDIR /usr/picard` and `RUN wget xxx/picard.jar` in Dockerfile and `java -Xmx2g -jar picard.jar FastqToSam` [Example Workflow](https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1/1d981269-88f5-466c-93d3-ab1d073279d9/timing)
5. [Not Work] `WORKDIR /usr/picard` and `ADD xxx/picard.jar /usr/picard/picard.jar` in Dockerfile and `java -Xmx2g -jar picard.jar FastqToSam` [Example Workflow](https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1/1d981269-88f5-466c-93d3-ab1d073279d9/timing)

docker/picard/Dockerfile
rexwangcc: Thanks! I will move it.

rexwangcc:
Now the picard docker image with R-dependencies has been pushed to the Docker Hub, please use `humancellatlas/picard:2.10.10-7ab16db` to use this image!

PR #52, opened 2017-12-06
Add analysis-tools Dockerfile.

rexwangcc:
In this PR I create a Dockerfile for analysis-tools, which would be useful for large-scale analysis as I discussed with @jishuxu . 

Note:
1. This docker image contains **Credentials**, so it should never be pushed to any public container registries such as Docker Hub or Quay. Currently, it is sitting on the gcr.io as `gcr.io/broad-dsde-mint-dev/analysis-tools:0.0.1`
2. This Dockerfile is a temporary solution for analysis and development, should be replaced by using credentials that are directly mounted from Vault in the future.


rexwangcc:
Going to merge it.

PR #61, opened 2017-12-18
Add gcloud components to analysis-tools docker and fix minor bugs.

rexwangcc:
In this PR, I add gcloud components to the docker image and make it authenticated at build time, so `gsutil` and other tools could be used in the container.

### DIFF
--- a/docker/analysis-tools/Dockerfile
+++ b/docker/analysis-tools/Dockerfile
@@ -36,10 +36,12 @@ RUN apt update && apt install -y \
   	libboost-all-dev \
 	libbz2-dev \
 	liblzma-dev \
+	lsb-release \
 	r-base \
   	r-base-core \
   	r-base-dev \
 	samtools \
+	sudo \
 	wget \
 	zlib1g-dev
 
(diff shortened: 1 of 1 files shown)


rexwangcc:
Sorry for the delay. @jishuxu PTAL

docker/analysis-tools/Dockerfile
```diff
@@ -68,14 +71,17 @@ RUN Rscript -e "install.packages('ggplot2')"
 RUN Rscript -e "install.packages('googleCloudStorageR')"
 
 # Mount Google Cloud secret key
-ADD broad-dsde-mint-dev-ed8cc09c48e5.json /usr/secrets/broad-dsde-mint-dev-ed8cc09c48e5.json
+ADD broad-dsde-mint-dev-testR.json /usr/secrets/broad-dsde-mint-dev-testR.json
```
jishuxu: are they the same json file as before but renamed?

rexwangcc: Yeah, they are the same, just renamed it to a more human-readable name.

PR #62, opened 2017-12-21
Ss2 tuneup pipelines

jishuxu:
In this PR
1) create hisat2 QC pipeline, include alignments and Picard runs 
2) add --calc-pme --single-cell-prior in rsem, which can incorporate with single cell data.
3) production pipeline wdls and example json files. Include hisat2 QC pipeline and hisat2rsem gene quantification

### DIFF
--- a/pipelines/tasks/hisat2.wdl
+++ b/pipelines/tasks/hisat2.wdl
@@ -1,4 +1,7 @@
 ## paired-end alignment
+## run HISAT2 to genome reference with dedault parameters
+## --seed to fix pseudo-random number and in order to produce deterministics results
+## -k --secondary to output multiple mapping reads. --keep 10 will output up to 10 multiple mapping reads, which is default in HISAT2
 task HISAT2PE {
   File hisat2_ref
   File fq1
@@ -6,7 +9,9 @@ task HISAT2PE {
   String ref_name
   String output_name
   String sample_name
+  Int disk_size
   command {
+    set -e
     tar -zxvf "${hisat2_ref}"
     hisat2 -t \
       -x ${ref_name}/${ref_name} \
@@ -16,14 +21,18 @@ task HISAT2PE {
       --rg PL:ILLUMINA --rg PU:${sample_name} \
       --new-summary --summary-file ${output_name}.log \
       --met-file ${output_name}.hisat2.met.txt --met 5 \
-      -p 4 -S ${output_name}.sam
+      --seed 12345 \
+      -k 10 \
+      --secondary \
+      -p 4 -S ${output_name}.sam 
     samtools sort -@ 4 -O bam -o "${output_name}.bam" "${output_name}.sam" 
   }
   runtime {
     docker:"quay.io/humancellatlas/secondary-analysis-hisat2:2-2.1.0"
     memory:"5 GB"
-    disks: "local-disk 100 HDD"
+    disks: "local-disk " + disk_size + " HDD"
     cpu: "4"
+    preemptible: 5
   }
   output {
     File logfile = "${output_name}.log"
@@ -33,8 +42,12 @@ task HISAT2PE {
 }
 ## run hisat2 for rsem
 ## increase gap alignment penalty to avoid gap alignment
-## --mp 1,1 --np 1 --score-min L,0,-0.1 is default paramesters when rsem runs alignment by using bowtie2
-## no-splice-alignment otherwise rsem will report partial alignment error.
+## --mp 1,1 --np 1 --score-min L,0,-0.1 is default paramesters when rsem runs alignment by using bowtie2/Bowtie
+## --mp 1,1 and --np 1 will reduce mismatching penalty to 1 for all.
+## with no-splice-alignment no-softclip no-mixed options on, HISAT2 will only output concordant alignment without soft-cliping
+## --rdg 99999999,99999999 and --rfg 99999999,99999999 will give an infinity penalty to alignment with indel.As results
+## no indel/gaps in alignments
+ 
 task HISAT2rsem {
   File hisat2_ref
   File fq1
@@ -42,7 +55,9 @@ task HISAT2rsem {
   String ref_name
   String output_name
   String sample_name
+  Int disk_size
   command {
+    set -e
     tar -zxvf "${hisat2_ref}" 
     hisat2 -t \
       -x ${ref_name}/${ref_name} \
@@ -53,6 +68,9 @@ task HISAT2rsem {
       --new-summary --summary-file ${output_name}.log \
       --met-file ${output_name}.hisat2.met.txt --met 5 \
       -k 10 \
+      --mp 1,1 \
+      --np 1 \
+      --score-min L,0,-0.1 \
       --secondary \
       --no-mixed \
       --no-softclip \
@@ -60,14 +78,16 @@ task HISAT2rsem {
       --rdg 99999999,99999999 \
       --rfg 99999999,99999999 \
       --no-spliced-alignment \
+      --seed 12345 \
       -p 4 -S ${output_name}.sam 
-    samtools view -bS "${output_name}.sam" >"${output_name}.bam"
+    samtools view -bS  "${output_name}.sam" > "${output_name}.bam"
   }
   runtime {
     docker:"quay.io/humancellatlas/secondary-analysis-hisat2:2-2.1.0"
     memory:"5 GB"
-    disks: "local-disk 50 HDD"
+    disks: "local-disk " + disk_size + " HDD"
     cpu: "4"
+    preemptible: 5
   }
   output {
     File logfile = "${output_name}.log"
@@ -82,7 +102,9 @@ task HISAT2SE {
   String ref_name
   String output_name
   String sample_name
+  Int disk_size
   command {
+    set -e
     tar -zxvf "${hisat2_ref}"
     hisat2 -t \
       -x ${ref_name}/${ref_name} \
@@ -91,14 +113,16 @@ task HISAT2SE {
       --rg PL:ILLUMINA --rg PU:${sample_name} \
       --new-summary --summary-file "${output_name}.log" \
       --met-file ${output_name}.hisat2.met.txt --met 5 \
+      --seed 12345 \
       -p 4 -S ${output_name}.sam
-      samtools sort -@ 4 -O bam -o "${output_name}.bam" "${output_name}.sam"
+    samtools sort -@ 4 -O bam -o "${output_name}.bam" "${output_name}.sam"
   }
   runtime {
     docker:"quay.io/humancellatlas/secondary-analysis-hisat2:2-2.1.0"
     memory:"5 GB"
-    disks: "local-disk 25 HDD"
+    disks: "local-disk " + disk_size + " HDD"
     cpu: "4"
+    preemptible: 5
   }
   output {
     File logfile ="${output_name}.log"
@@ -110,8 +134,9 @@ task HISAT2SE {
 task hisat2_inspect_index {
   File hisat2_ref
   String ref_name
-
+  Int disk_size
   command {
+    set -e
     tar -zxvf "${hisat2_ref}"
     hisat2-inspect --ss --snp \
        -s ${ref_name}/${ref_name} > hisat2_inspect.log
@@ -119,8 +144,9 @@ task hisat2_inspect_index {
   runtime {
     docker:"quay.io/humancellatlas/secondary-analysis-hisat2:2-2.1.0"
     memory:"3 GB"
-    disks: "local-disk 25 HDD"
+    disks: "local-disk " + disk_size + " HDD"
     cpu: "1"
+    preemptible: 5
   }
   output {
     File logfile ="hisat2_inspect.log"
--- a/pipelines/tasks/picard.wdl
+++ b/pipelines/tasks/picard.wdl
@@ -2,7 +2,7 @@ task CollectMultipleMetrics {
   File aligned_bam
   File ref_genome_fasta
   String output_filename
-  
+  Int disk_size 
   command {
     java -Xmx6g -jar /usr/picard/picard.jar CollectMultipleMetrics \
       VALIDATION_STRINGENCY=SILENT \
@@ -25,7 +25,8 @@ task CollectMultipleMetrics {
   runtime {
     docker:"quay.io/humancellatlas/secondary-analysis-picard:2.10.10-7ab16db"
     memory:"7.5 GB"
-    disks: "local-disk 50 HDD"
+    disks: "local-disk " + disk_size + " HDD"
+    preemptible: 5
   }
   output {
     File alignment_summary_metrics = "${output_filename}.alignment_summary_metrics.txt"
@@ -54,7 +55,9 @@ task CollectRnaMetrics {
   File rrna_interval
   String output_filename
   String stranded
+  Int disk_size
   command{
+    set -e
     java -Xmx3g -jar /usr/picard/picard.jar CollectRnaSeqMetrics \
       VALIDATION_STRINGENCY=SILENT \
       METRIC_ACCUMULATION_LEVEL=ALL_READS \
@@ -64,24 +67,27 @@ task CollectRnaMetrics {
       RIBOSOMAL_INTERVALS="${rrna_interval}" \
       STRAND_SPECIFICITY=${stranded} \
       CHART_OUTPUT="${output_filename}.rna.coverage.pdf"
+    touch "${output_filename}.rna.coverage.pdf"
   }
   runtime {
     docker:"quay.io/humancellatlas/secondary-analysis-picard:2.10.10-7ab16db"
     memory:"3.75 GB"
-    disks: "local-disk 10 HDD"
+    disks: "local-disk " + disk_size + " HDD"
+    preemptible: 5
   }
   output {
     File rna_metrics = "${output_filename}.rna_metrics.txt"
-    File rna_coverage = "${output_filename}.rna.coverage.pdf"
+    File rna_coverage_pdf = "${output_filename}.rna.coverage.pdf"
   }
 }
-
+## Here are use  -XX:ParallelGCThreads=2 to run MarkDuplication on mutlple
+## thread. 
 task CollectDuplicationMetrics {
   File aligned_bam
   String output_filename
-
+  Int disk_size
   command {
-    java -Xmx6g -jar /usr/picard/picard.jar  MarkDuplicates \
+    java -Xmx6g -XX:ParallelGCThreads=2  -jar /usr/picard/picard.jar  MarkDuplicates \
        VALIDATION_STRINGENCY=SILENT  \
        INPUT=${aligned_bam} \
        OUTPUT="${output_filename}.MarkDuplicated.bam" \
@@ -92,11 +98,12 @@ task CollectDuplicationMetrics {
   runtime {
     docker: "quay.io/humancellatlas/secondary-analysis-picard:2.10.10-7ab16db"
     memory: "7.5 GB"
-    disks: "local-disk 50 HDD"
+    cpu: "2"
+    disks: "local-disk " + disk_size + " HDD"
+    preemptible: 5
   }
   output {
     File dedup_metrics = "${output_filename}.duplicate_metrics.txt"
-    File dedup_bamfile = "${output_filename}.MarkDuplicated.bam"
   }
 }
 
--- a/pipelines/tasks/rsem.wdl
+++ b/pipelines/tasks/rsem.wdl
@@ -1,25 +1,43 @@
+## RSEM will estimate gene/isoform quantification
+## --bam: input is bam file
+## -p 4: run on multiple threads
+## --time: report running time
+## --seed: report deterministic results
+## --calc-pme will do posterior mean estimation
+## --single-cell-prior in pme estimation, use Dirichlet(0.1) as the prior 
+## which encourage the sparsity of the expression levels
+## of note, the --single-cell-prior has not been tested yet and there are
+## more investigation or improvement requested.  
 task RsemExpression {
   File trans_aligned_bam
   File rsem_genome
   String rsem_out
-  
+  Int disk_size
   command {
+    set -e
+  
     tar -xvf ${rsem_genome}
-    echo "Aligning fastqs and calculating expression"
-    rsem-calculate-expression --bam --paired-end ${trans_aligned_bam} rsem/rsem_trans_index  "${rsem_out}" -p 4 --ci-memory 2000 --time 
-    ## parse gene expected_count out
-    cut -f 1,4,5 "${rsem_out}.genes.results" >"${rsem_out}.gene.expected_counts"
+    rsem-calculate-expression \
+      --bam \
+      --paired-end \
+       -p 4 \
+      --time --seed 555 \
+      --calc-pme \
+      --single-cell-prior \
+      ${trans_aligned_bam} \
+      rsem/rsem_trans_index  \
+      "${rsem_out}" 
   }
   runtime {
     docker: "quay.io/humancellatlas/secondary-analysis-rsem:1.3.0"
     memory: "3.75 GB"
-    disks: "local-disk 50 HDD"
+    disks: "local-disk " + disk_size + " HDD"
     cpu: "4"
+    preemptible: 5
   }
   output {
     File rsem_gene = "${rsem_out}.genes.results"
     File rsem_isoform = "${rsem_out}.isoforms.results"
-    File rsem_gene_count = "${rsem_out}.gene.expected_counts"
     File rsem_time = "${rsem_out}.time"
     File rsem_cnt = "${rsem_out}.stat/${rsem_out}.cnt"
     File rsem_model = "${rsem_out}.stat/${rsem_out}.model"
--- a/smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl
+++ b/smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl
@@ -1,0 +1,88 @@
+import "hisat2.wdl" as hisat2
+import "picard.wdl" as picard
+## QC pipeline
+## HISAT2 as aligner to align reads to genome reference
+## output: genome reference aligned bam file 
+## Picard will produce a set of QC metricss
+## output: a set of metrics files. 
+workflow RunHisat2Pipeline {
+  File fastq_read1
+  File fastq_read2
+  File gtf
+  File stranded
+  File ref_fasta
+  File rrna_interval
+  File ref_flat
+  File hisat2_ref
+  String output_prefix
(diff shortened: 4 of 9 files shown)


pipelines/tasks/hisat2.wdl
```diff
@@ -16,7 +16,10 @@ task HISAT2PE {
       --rg PL:ILLUMINA --rg PU:${sample_name} \
       --new-summary --summary-file ${output_name}.log \
       --met-file ${output_name}.hisat2.met.txt --met 5 \
-      -p 4 -S ${output_name}.sam
+      --seed 12345 \
+      -k 10 \
+      --secondary \
+      -p 4 -S ${output_name}.sam
```
ambrosejcarr: If they really do have only short-name parameters, I wouldn't mind having some comments about why each parameter is chosen (in cases where you chose them because "these are suggested to work with RSEM", that's enough of a comment).

pipelines/tasks/hisat2.wdl
```diff
@@ -53,15 +56,19 @@ task HISAT2rsem {
       --new-summary --summary-file ${output_name}.log \
       --met-file ${output_name}.hisat2.met.txt --met 5 \
       -k 10 \
+      --mp 1,1 \
```
ambrosejcarr: Same comment here; please explain parameter choices. Ideally I should be able to pick up your task wdls and understand how to use them and what they do (beyond just "aligns reads").

pipelines/tasks/picard.wdl
```diff
@@ -81,7 +81,7 @@ task CollectDuplicationMetrics {
   String output_filename
 
   command {
-    java -Xmx6g -jar /usr/picard/picard.jar  MarkDuplicates \
+    java -Xmx6g -XX:ParallelGCThreads=2 -jar /usr/picard/picard.jar  MarkDuplicates \
```
ambrosejcarr: Did this speed things up?

pipelines/tasks/rsem.wdl
```diff
@@ -19,7 +19,7 @@ task RsemExpression {
   output {
     File rsem_gene = "${rsem_out}.genes.results"
     File rsem_isoform = "${rsem_out}.isoforms.results"
-    File rsem_gene_count = "${rsem_out}.gene.expected_counts"
+    ## File rsem_gene_count = "${rsem_out}.gene.expected_counts"
```
ambrosejcarr: Why did you choose to drop this? What was it previously doing?

smartseq2_single_sample/ss2_single_sample.wdl
```diff
@@ -1,192 +1,50 @@
...
-  File aligned_bam
-  File gtf
-  String fc_out
+import "hisat2_QC_pipeline.wdl" as run_hisat2
+import "hisat2_rsem_pipeline.wdl" as run_hisat2_rsem
+
+workflow run_pipeline {
+
+  # load annotation
+  File gtffile
```
ambrosejcarr: this parameter should have an underscore

smartseq2_single_sample/ss2_single_sample.wdl
```diff
@@ -1,192 +1,50 @@
...
-    memory: "40 GB"
-    disks :"local-disk 100 HDD"
-  }
-}
-
-task FeatureCountsUniqueMapping {
-  File aligned_bam
-  File gtf
-  String fc_out
+import "hisat2_QC_pipeline.wdl" as run_hisat2
```
ambrosejcarr: Please add more comments about why you are running hisat2_QC and also hisat2_rsem -- I doubt this is transparent to users.

pipelines/tasks/rsem.wdl
kbergin: If it's really being dropped, should it be taken out vs commented out?

pipelines/tasks/hisat2.wdl
kbergin: If they're short name parameters it may be helpful to have some comments explaining what they are.. or use the long names if that's easier?

kbergin: It won't let me comment below, but do we have a policy on how we want the formatting to be with respect to maybe white space between task definitions? I like having one enter between them personally

kbergin: And do we want to have comments about parameter choices next to the parameter inline instead of above? Just a thought

pipelines/tasks/rsem.wdl
```diff
@@ -6,9 +6,9 @@ task RsemExpression {
   command {
     tar -xvf ${rsem_genome}
     echo "Aligning fastqs and calculating expression"
-    rsem-calculate-expression --bam --paired-end ${trans_aligned_bam} rsem/rsem_trans_index  "${rsem_out}" -p 4 --ci-memory 2000 --time 
+    rsem-calculate-expression --bam --paired-end ${trans_aligned_bam} rsem/rsem_trans_index  "${rsem_out}" -p 4 --calc-pme --single-cell-prior --time --seed 555
     ## parse gene expected_count out
-    cut -f 1,4,5 "${rsem_out}.genes.results" >"${rsem_out}.gene.expected_counts"
+    ## cut -f 1,4,5 "${rsem_out}.genes.results" >"${rsem_out}.gene.expected_counts"
```
kbergin: If this is being taken out, we should remove it, not comment it out. Including removing the comment above it

pipelines/tasks/hisat2.wdl
jsotobroad: we should be consistent one way or the other, the `picard.wdl` has a `\n` between tasks.  Spacing between task input declaration and different blocks in the task itself are worth making consistent as well.

pipelines/tasks/picard.wdl
jsotobroad: This would be handy for greenteam if it did!  Maybe more importantly did it make it cheaper?

pipelines/tasks/picard.wdl
```diff
@@ -92,6 +92,7 @@ task CollectDuplicationMetrics {
   runtime {
     docker: "quay.io/humancellatlas/secondary-analysis-picard:2.10.10-7ab16db"
     memory: "7.5 GB"
+    cpu: "2"
     disks: "local-disk 50 HDD"
```
jsotobroad: Hardcoded disk sizes scare me when it comes to production wdls.  We should think about dynamically sizing our wdls (not sure if now is the right/best time but in general).  There are a number of examples in other wdls, https://github.com/broadinstitute/dsde-pipelines/blob/develop/genomes_in_the_cloud/single_sample/PairedSingleSampleWf.wdl#L145 has a couple of them.  This helps with both cost and robustness of the workflow

pipelines/tasks/hisat2.wdl
```diff
@@ -53,15 +56,19 @@ task HISAT2rsem {
...
       --no-mixed \
       --no-softclip \
       --no-discordant \
       --rdg 99999999,99999999 \
       --rfg 99999999,99999999 \
       --no-spliced-alignment \
+      --seed 12345 \
       -p 4 -S ${output_name}.sam 
-    samtools view -bS "${output_name}.sam" >"${output_name}.bam"
+    samtools view -bS  "${output_name}.sam" >"${output_name}.bam"
```
jsotobroad: seems like an extra space snuck in here?

pipelines/tasks/hisat2.wdl
```diff
@@ -16,7 +16,10 @@ task HISAT2PE {
       --rg PL:ILLUMINA --rg PU:${sample_name} \
       --new-summary --summary-file ${output_name}.log \
       --met-file ${output_name}.hisat2.met.txt --met 5 \
-      -p 4 -S ${output_name}.sam
+      --seed 12345 \
```
jsotobroad: you want to add `set -e` to the top any command block that runs multiple commands.  This will let you catch any intermediate errors. https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html.  You could also end each of your commands with `&&` instead (except the last one)

pipelines/tasks/hisat2.wdl
```diff
@@ -53,15 +56,19 @@ task HISAT2rsem {
       --new-summary --summary-file ${output_name}.log \
```
jsotobroad: same `set -e` comment

pipelines/tasks/rsem.wdl
```diff
@@ -6,9 +6,9 @@ task RsemExpression {
   command {
     tar -xvf ${rsem_genome}
```
jsotobroad: `set -e`

jishuxu:
I add some comments to explain parameters choice in wdl. let me know if  it is not clear. I also change the disk size to be estimated based on inputs. But the choice is just arbitrary, we may do more to optimize them later.

pipelines/tasks/hisat2.wdl
jishuxu: added

pipelines/tasks/hisat2.wdl
jsotobroad: sorry nitpicky space here

smartseq2_single_sample/pipelines/hisat2_rsem_pipeline.wdl
```diff
@@ -9,6 +14,14 @@ workflow RunHisat2RsemPipeline {
   String output_prefix
   String hisat2_ref_trans_name
   String sample_name
+  ##variable to estimate disk size
+  ## variables to estimate disk size
+  Float hisat2_ref_size = size(hisat2_ref_trans,"GB")
```
jsotobroad: can you add a space after the `,`s

smartseq2_single_sample/pipelines/hisat2_rsem_pipeline.wdl
```diff
@@ -17,25 +30,27 @@ workflow RunHisat2RsemPipeline {
       fq2 = fastq_read2,
       ref_name = hisat2_ref_trans_name,
       sample_name = sample_name,
-      output_name = output_prefix
+      output_name = output_prefix,
+      disk_size = fastq_size*md_disk_multiplier+hisat2_ref_size+additional_disk*5.0
```
jsotobroad: As a personal preference, I prefer spaces before/after operators.  I think its dealer's choice here though.

smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl
```diff
@@ -0,0 +1,87 @@
...
+ 
+  call hisat2.HISAT2PE as Hisat2 {
+    input:
+      hisat2_ref = hisat2_ref,
+      fq1 = fastq_read1,
+      fq2 = fastq_read2,
+      ref_name = hisat2_ref_name,
+      sample_name = sample_name,
+      output_name = output_prefix,
+      disk_size = hisat2_ref_size+fastq_size*md_disk_multiplier+additional_disk*5.0
```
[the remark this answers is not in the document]
ambrosejcarr: I would also be consistent about whitespace here; I like spaces around operators. It also seems odd that "additional_disk" gets multiplied by 5; shouldn't the user get to directly specify the additional disk size without modification? 

Further comments: 
- the calculation seems quite complicated. What does `md` refer to in `md_disk_multiplier`? Is it designed to just multiply the fastq size as coded?

smartseq2_single_sample/pipelines/hisat2_rsem_pipeline.wdl
```diff
@@ -31,7 +31,7 @@ workflow RunHisat2RsemPipeline {
       ref_name = hisat2_ref_trans_name,
       sample_name = sample_name,
       output_name = output_prefix,
-      disk_size = fastq_size*md_disk_multiplier+hisat2_ref_size+additional_disk
+      disk_size = fastq_size*md_disk_multiplier+hisat2_ref_size+additional_disk*5.0
```
ambrosejcarr: Same as above. Also, is disk size specified in GB by default? I thought it was bytes, in which case this may not function well with the `size` calls above. 

... the wdl draft 1 and 2 specs don't seem to comment on `disk_size` at all, so I don't know the answer for sure. You could check in a metadata file.

smartseq2_single_sample/pipelines/run_pipeline_scatter.wdl
```diff
@@ -0,0 +1,53 @@
+import "hisat2_QC_pipeline.wdl" as run_hisat2
+import "hisat2_rsem_pipeline.wdl" as run_hisat2_rsem
+
+workflow run_pipelines {
+
+  # load annotation
+  File gtffile
```
ambrosejcarr: `gtffile` should have a space here (--> `gtf_file`) to be consistent with other names, as should `sampleinfo` below.

pipelines/tasks/hisat2.wdl
```diff
@@ -33,16 +44,23 @@ task HISAT2PE {
 }
 ## run hisat2 for rsem
 ## increase gap alignment penalty to avoid gap alignment
-## --mp 1,1 --np 1 --score-min L,0,-0.1 is default paramesters when rsem runs alignment by using bowtie2
-## no-splice-alignment otherwise rsem will report partial alignment error.
+## --mp 1,1 --np 1 --score-min L,0,-0.1 is default paramesters when rsem runs alignment by using bowtie2/Bowtie
+## --mp 1,1 and --np 1 will reduce mismatching penalty to 1 for all.
+## with no-splice-alignment no-softclip no-mixed options on, HISAT2 will onlu output concordance alignment
+## --rdg 99999999,99999999 and --rfg 99999999,99999999 will give an infinity penalty to alignment with indel.As results
```
ambrosejcarr: Small typos here; space after period, not sure if the subsequent line is a continuation of this one, but if it is it should be indented to be clear.

pipelines/tasks/hisat2.wdl
```diff
@@ -1,12 +1,19 @@
 ## paired-end alignment
+## run HISAT2 to genome reference with dedault parameters
+## --seed to fix pseudo-random number and in order to produce deterministics results
+## -k --secondary to output multiple mapping reads. --keep 10 will output  up to 10 multiple mapping reads
+## which actually is default
```
ambrosejcarr: indent "which" to make clear it continues from previous line.

pipelines/tasks/hisat2.wdl
```diff
@@ -33,16 +44,23 @@ task HISAT2PE {
 }
 ## run hisat2 for rsem
 ## increase gap alignment penalty to avoid gap alignment
-## --mp 1,1 --np 1 --score-min L,0,-0.1 is default paramesters when rsem runs alignment by using bowtie2
-## no-splice-alignment otherwise rsem will report partial alignment error.
+## --mp 1,1 --np 1 --score-min L,0,-0.1 is default paramesters when rsem runs alignment by using bowtie2/Bowtie
+## --mp 1,1 and --np 1 will reduce mismatching penalty to 1 for all.
+## with no-splice-alignment no-softclip no-mixed options on, HISAT2 will onlu output concordance alignment
```
ambrosejcarr: onlu -> only, concordance -> concordant

pipelines/tasks/rsem.wdl
```diff
@@ -1,25 +1,43 @@
+## RSEM will estimate gene/isoform quantification
+## --bam: input is bam file
+## -p 4: run on multiple threads
+## --time: report running time
+## --seed: report deterministic results
+## --calc-pme will do posterior mean estimation
+## --single-cell-prior in pme estimation, use Dirichlet(0.1) as the prior 
+## which encourage the sparsity of the expression levels
```
ambrosejcarr: indent line (and subsequent)

smartseq2_single_sample/archived/ss2_single_sample.wdl
```diff
@@ -0,0 +1,192 @@
...
+  File trans_aligned_bam
+  File rsem_genome
+  String rsem_out
+  
+  command {
+    tar -xvf ${rsem_genome}
+    echo "Aligning fastqs and calculating expression"
+    rsem-calculate-expression --bam --paired-end ${trans_aligned_bam} rsem/rsem_trans_index  "${rsem_out}"
+    ## parse gene expected_count out
+    cut -f 1,4,5 "${rsem_out}.genes.results" >"${rsem_out}.gene.expected_counts"
```
ambrosejcarr: space after operator.

smartseq2_single_sample/archived/ss2_single_sample.wdl
```diff
@@ -0,0 +1,192 @@
...
+    input:
+      input_fastq_read1 = fastq_read1,
+      input_fastq_read2 = fastq_read2,
+      gtf = gtf,
+      star_genome = star_genome
+  }
+ 
+  call RsemExpression {
+    input:
+      trans_aligned_bam=Star.output_bam_trans,
```
ambrosejcarr: Be consistent about `name=object` or `name = object`

smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl
```diff
@@ -0,0 +1,87 @@
+import "hisat2.wdl" as hisat2
+import "picard.wdl" as picard
+## QC pipeline
```
ambrosejcarr: I think you should be a bit more verbose about why you do this as opposed to explaining what is done, because it seems a bit crazy to run alignment twice without understanding that we need to do it for consistency with picard's expectations about BAM formats. 

It also has the potential to help engineers who might have ideas about how to fix this problem that we haven't thought about.

smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl
```diff
@@ -0,0 +1,87 @@
...
+  }
+  call picard.CollectDuplicationMetrics {
+    input:
+      aligned_bam = Hisat2.output_bam,
+      output_filename = output_prefix,
+      disk_size = (md_disk_multiplier * bam_size) + additional_disk
+  }
+
+  output {
+    File aligned_bam = Hisat2.output_bam
```
ambrosejcarr: Be consistent here with whitespace.`gc_bias_detail_metrics` has multiple spaces, `rna_coverage` is missing a space.

smartseq2_single_sample/pipelines/hisat2_rsem_pipeline.wdl
ambrosejcarr: Seconded.

smartseq2_single_sample/ss2_single_sample.wdl
```diff
@@ -1,192 +1,53 @@
...
-  }
-
-  call FeatureCountsUniqueMapping {
+import "hisat2_QC_pipeline.wdl" as run_hisat2
+import "hisat2_rsem_pipeline.wdl" as run_hisat2_rsem
+## main pipeline:
+## QC pipeline: HISAT2+Picard on genome alignments
+## quantification pipeline: HISAT2+RSEM on transcriptome alignments
+
+workflow run_pipeline {
```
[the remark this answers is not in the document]
jishuxu: yes, just changes

jishuxu:
some updates
1) remove 'archived' folder which was just a placeholder for previous pipeline.
2) remove rna_coverage pdf output from Picard rna metrics collection. In current Picard RNA metrics module, this pdf will be plotted only if there is actual sufficient RNA coverage, it will be troublesome if  single cell samples are at very low quality, no RNA coverage.
3) remove marked duplication bam file from output. we actual never use this bam in analysis.

smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl
```diff
@@ -17,8 +17,8 @@ workflow RunHisat2Pipeline {
   String sample_name
   ## variables to estimate disk size
   Float hisat2_ref_size = size(hisat2_ref,"GB")
-  Float fastq_size = size(fastq_read1,"GB") +size(fastq_read2,"GB")
-  Float reference_bundle_size = size(ref_fasta,"GB") + size(ref_flat,"GB")+size(rrna_interval,"GB")+size(gtf,"GB")
+  Float fastq_size = size(fastq_read1,"GB") + size(fastq_read2,"GB")
+  Float reference_bundle_size = size(ref_fasta,"GB") + size(ref_flat,"GB") + size(rrna_interval,"GB") + size(gtf,"GB")
   Float md_disk_multiplier = 10.0
   Int? increase_disk_size
```
ambrosejcarr: This `increase_disk_size` still confuses me. Could you explain what you're trying to accomplish to me? Maybe I just don't get it.

smartseq2_single_sample/ss2_single_sample.wdl
```diff
@@ -4,7 +4,7 @@ import "hisat2_rsem_pipeline.wdl" as run_hisat2_rsem
 ## QC pipeline: HISAT2+Picard on genome alignments
 ## quantification pipeline: HISAT2+RSEM on transcriptome alignments
 
-workflow run_pipeline {
+workflow Ss2RsemSingleSample {
```
ambrosejcarr: I like the change, but could you include Hisat2 in the name as well to be more descriptive, and since that has changed since the last iteration? @dshiga Would my suggestion be consistent with your request?

dshiga: Now that you mention it, do we really need Rsem in there? Why not just Ss2SingleSample or Ss2 for that matter? Do we expect to have multiple ss2 pipelines, some using rsem and hisat, and others using other aligners?

ambrosejcarr: My preference if I could name the production pipeline, would be for `SmartSeq2`as I think that most accurately encapsulates the data type we expect to process with this pipeline. 

There is the trickiness that technically this pipeline only runs on one cell and the _adapter_ scatters over the cells. For this reason I would also accept `OneCellSmartSeq2` or similar. I don't think `SingleSample` conveys this information, as many might think that experiment equates to sample. Similarly, I think `SingleCell` is overloaded in this context. One would naturally assume we refer to single-cell sequencing.

ambrosejcarr: ... corollary: If we go with `SmartSeq2`, we should put a note into the workflow about dependence upon the adapter to process a complete multi-cell experiment.

jishuxu:
about the disk_size question, here are some details as following. 
Besides the `fastq files size`, `reference size`, there are two more parameters, `md_disk_multiplier` and `additional_disk`
1) `md_disk_multiplier` is used to estimated the aligned bam file size. I gave "10" is mainly because I allow up to 10 multiple mapping in alignment. 
2) `additional_disk` is added here to give more space. For example, hisat2 task includes using samtools to sort sam file, so I gave `additional_disk*5.0`. and rsem task does not include sorting but only converting sam to bam, so I just gave `additional_disk*2.0`. 
3) the default or initial value of `additional_disk` is `10`, it is just an empirical value. The `50Gb` was the previous disk space I gave to HISAT2 alignment task. Most time the 50Gb disk space is sufficient. but in some case, let's say 6 out of 860 times, it did not work and required more disk space.

jishuxu:
After talked with @jsotobroad , I decided to add rna_coverage_pdf back to outputs and added cmd line `touch rna_coverage_pdf` to avoid the error.

smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl
jishuxu: added some comments in other wdl file and hope it will help

ambrosejcarr:
Ok, with respect to the disk multiplier, I think that makes sense to parameterize since it depends on multiple alignments. Please change `md` to `multimapping` or `multiple_mapping` or something that makes it clear what it refers to, and make a comment the first time it's used. 

For `additional_disk`, it sounds like each task requires a different value based on its specific contents. As such,  I don't think this one makes sense to parameterize. I would just fold it into each task with `+x` where `x` is the number of GB that a particular task needs in excess of the input files. Alternatively, you can have task-specific multipliers if you expected to need to sort a bam file, you could have `x 2`.

smartseq2_single_sample/ss2_single_sample.wdl
jishuxu: I think we should use a name not too specific but also descriptive, I would call it something like `Ss2Hisat2RsemPipeline` and a wrapper which scatter-gather this pipeline would be called something like `Ss2Hisat2RsemPipelineWrapper`

jishuxu:
I rename `md_disk_multiplier` to `bam_disk_multiplier`. it is really a bam file size estimator. As for the `additional_disk`, I would keep as it is for now and we can change it later when we optimize the pipeline.

pipelines/tasks/hisat2.wdl
```diff
@@ -16,14 +21,18 @@ task HISAT2PE {
...
+      -p 4 -S ${output_name}.sam 
     samtools sort -@ 4 -O bam -o "${output_name}.bam" "${output_name}.sam" 
   }
   runtime {
     docker:"quay.io/humancellatlas/secondary-analysis-hisat2:2-2.1.0"
     memory:"5 GB"
-    disks: "local-disk 100 HDD"
+    disks: "local-disk " + disk_size + " HDD"
     cpu: "4"
+    preemptible: 5
```
jsotobroad: gonna be so cheap!

jishuxu: Yes, i will re-run a full dataset after this merged and we can figure out how cheap it will be!

jishuxu:
I changed the pipeline name to `Ss2SingleCellPipeline`. @dshiga @ambrosejcarr  is it okay with you?

dshiga:
Works for me

ambrosejcarr:
Works for me. I'd rather `SmartSeq2SingleCellPipeline` -- may as well remove the acronym?

dshiga:
One more thought. Is "pipeline" necessary or is it implied? I'm fine with
it either way or Ambrose's way.

smartseq2_single_sample/pipelines/run_pipeline_scatter.wdl
```diff
@@ -0,0 +1,53 @@
...
+  String stranded
+  File sample_info
+  String sra_dir
+  Array[String] sraIDs = read_lines(sample_info)
+  
+  scatter(idx in range(length(sraIDs))) { 
+    call run_hisat2.RunHisat2Pipeline {
+      input:
+        fastq_read1 = sra_dir+'/'+sraIDs[idx]+"_1.fastq.gz",
+        fastq_read2 = sra_dir+'/'+sraIDs[idx]+"_2.fastq.gz",
```
mckinsel: This seems pretty tightly coupled to a particular bucket. How would this work outside of the Broad infrastructure?

kbergin: I thiiiink this is a wdl just to run the test pipeline for analysis, so maybe not as concerning that it's portable. But fair point.

PR #65, opened 2018-01-04
Ajc fix hisat2 naming

ambrosejcarr:
Hisat2 requires files have .fq, .fastq, .fq.gz or .fastq.gz extensions. This PR implements a change to Hisat2 to rename `uuid` files to `uuid.fastq.gz`

I'm putting this up for feedback; right now it is quite fragile, and relies upon the fact that all files in blue box are .fastq.gz format/compressed.

### DIFF
--- a/pipelines/tasks/hisat2.wdl
+++ b/pipelines/tasks/hisat2.wdl
@@ -4,19 +4,38 @@
 ## -k --secondary to output multiple mapping reads. --keep 10 will output up to 10 multiple mapping reads, which is default in HISAT2
 task HISAT2PE {
   File hisat2_ref
-  File fq1
-  File fq2
+  File fq1  # gz forward fastq file
+  File fq2  # gz reverse fastq file
   String ref_name
   String output_name
   String sample_name
   Int disk_size
   command {
+    # Note that files MUST be gzipped or the module will not function properly
+    # This will be addressed in the future either by a change in how Hisat2 functions or a more
+    # robust test for compression type.
+
     set -e
-    tar -zxvf "${hisat2_ref}"
+
+    # fix names if necessary.
+    if [ "${fq1}" != *.fastq.gz ]; then
+        FQ1=${fq1}.fastq.gz
+        mv ${fq1} ${fq1}.fastq.gz
+    else
+        FQ1=${fq1}
+    fi
+    if [ "${fq2}" != *.fastq.gz ]; then
+        FQ2=${fq2}.fastq.gz
+        mv ${fq2} ${fq2}.fastq.gz
+    else
+        FQ2=${fq2}
+    fi
+
+   tar -zxvf "${hisat2_ref}"
     hisat2 -t \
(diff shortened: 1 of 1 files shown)


jishuxu:
I just did test, interestingly, HISAT2 only requires files with validate extension, such as `gz`, `bz2`, the `fastq` or `fq` is not necessary in extension!  It is not what I expected!!!

kbergin:
Wouldn't it be possible to utilize the file type stored in the metadata in the blue box to correctly rename the file with it's extension in the adapter wdl instead of here in the hisat wdl?

jishuxu:
@kbergin  the issue here is really about HISAT2, hisat2 only take files with validated file extension, such as `gz` or `bz2`. The input files which our pipeline receives from blue box are all named by uuid without file extension. Ambrose tried to do here is just to add a file extension to the input so that HISAT2 won't report error.  There is an some improvement need to do but the approach is correct.

kbergin:
@jishuxu Maybe this is the only time thus far that we've run into this issue, but in general a lot of tools aren't set up to work with UUIDs without filetypes. I think a team discussion is needed on a long term plan here, but am open to this as a short term solution

jishuxu:
@kbergin, this issue might a common one for many of our pipeline production in the future. We should look into this. I actually had a discussion with Ambrose this morning and we tried to figure out the best feasible way to handle it. The current solution is just a quick fix which can allow Ambrose to work on the other issue with the wrapper wdl. The ideal solution is either we retrieve file format from metadata then add the correct file extension to uuid, or we validate file type on our own then add correct file extension to uuid. Either way is straight forward. I probably prefer to the second way.

mckinsel:
This is indeed a pretty general issue. One thing to consider is that tools can depend on the whole file name, not just the extension. There are unfortunately certain files that convey information in their name, most commonly Illumina fastq files produced by bcl2fastq.

kbergin:
I've made a ticket to reflect a longer term decision/discussion on this topic. If the team is fine with this short term fix I am as well. https://elastc.com/c/Ky6CfJeB/478-handling-uuids-without-file-extensions-from-blue

optimus/test_optimus.ipynb
```diff
@@ -404,13 +402,58 @@
...
+    "cromwell_server.query(status=['Running'], names=['Optimus']).json()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "metadata": {},
+   "outputs": [
+    {
+     "ename": "NameError",
```
dshiga: Does it make sense that we're checking in a version of the notebook with an error in it? I confess I don't really know what I'm looking at here. Is there a way to get rid of this error?

ambrosejcarr: Sorry, ignore the notebook; it will be removed before I merge. I didn't realize I had checked that in.

ambrosejcarr:
Will remove notebook and make a note that files must be gzipped. Thanks @dshiga

PR #68, opened 2018-01-09
Add a portability test of SmartSeq2 and Optimus

mckinsel:
This is not necessarily for merging (though it could be merged!)

This is a portability test of the SmartSeq2 pipeline. It uses travis to submit a portability test workflow to an externally-hosted portability testing service. That service in turn distributes the test workflow to different environments and reports on the success or failure of the workflow in those environments. The service code is in www.github.com/HumanCellAtlas/portability.

There are a few points here to for the Mint team to consider:
1. Skylab is being reorganized, how should portability test code fit with that?
2. This PR puts the tests in a json file with a (hopefully) straightforward structure. Does that work for defining the tests?
3. The current SS2 test finishes in ~20 minutes even on a 2 core machine. How long will future portability tests take? This will determine when/if we move off of travis, but more importantly it determines how these tests fit with developer practices. 
4. How do we host test input files? This involved me copying them to AWS and making them public. But that's not a real solution.

### DIFF
--- a/test/run_portability_test.py
+++ b/test/run_portability_test.py
@@ -1,0 +1,199 @@
+"""Runs a single portability test from a test directory."""
+
+import argparse
+import datetime
+import glob
+import json
+import os
+import sys
+import time
+
+import requests
+
+# This is what we need to find in a test directory. I guess we'll glob the PR
+# test WDL itself since it's named after the pipeline.
+TEST_DIR_LAYOUT = {
+    "inputs": "test_inputs.json",
+    "dependencies": "dependencies.json",
+    "test": "*PR.wdl"
+}
+
+class TestFailure(Exception):
+    """Error for a test that fails its portability test."""
+    pass
+
+class TestDefinitionError(Exception):
+    """Error for malformed test directory."""
+    pass
+
+class TestEnvironmentError(Exception):
+    """Error for malformed test environment."""
+    pass
+
+def gather_test_inputs(test_dir):
+    """Walk through the test directory, finding files and dependencies needed
+    to submit a portability test.
+
+    Raise a TestDefinitionError if anything goes wrong.
+    """
+
+    errors = []
+
+    # Test WDL
+    wdl_glob = os.path.join(test_dir, TEST_DIR_LAYOUT["test"])
+    wdl_paths = glob.glob(wdl_glob)
+    if not wdl_paths:
+        errors.append("Test definition WDL not found.")
+    if len(wdl_paths) > 1:
+        errors.append("Multiple candidate test WDLs found: {}".format(wdl_paths))
+
+    try:
+        test_wdl_string = open(wdl_paths[0]).read()
+    except IOError:
+        test_wdl_string = None
+        errors.append("Test WDL {} could not be read".format(wdl_paths[0]))
+
+    # Test inputs
+    try:
+        inputs_json_path = os.path.join(test_dir, TEST_DIR_LAYOUT["inputs"])
+        inputs_json_string = open(inputs_json_path).read()
+    except IOError:
+        inputs_json_string = None
+        errors.append("Inputs JSON {} could not be read".format(inputs_json_path))
+
+    # Dependencies
+
+    # First try to load the dependency JSON itself
+    try:
+        dependencies_json_path = os.path.join(test_dir, TEST_DIR_LAYOUT["dependencies"])
+        dependencies_json_string = open(dependencies_json_path).read()
+    except IOError:
+        dependencies_json_string = None
+        errors.append("Dependencies JSON {} could not be read".format(dependencies_json_path))
+
+    # Now iterate over the dependencies and load the files.
+    loaded_dependencies = []
+    dependencies_dict = json.loads(dependencies_json_string)
+
+    for key, value in dependencies_dict.items():
+        try:
+            loaded_dependencies.append({
+                "name": key,
+                "code": open(value).read()
+            })
+        except IOError:
+            errors.append("Could not read dependency {}".format(value))
+
+    if errors:
+        for error in errors:
+            sys.stderr.write(error + "\n")
+        raise TestDefinitionError(errors)
+
+    return {
+        "workflow_descriptor": test_wdl_string,
+        "workflow_params": inputs_json_string,
+        "workflow_dependencies": loaded_dependencies
+        }
+
+# These are the environment variables we're expecting to be able to submit a
+# portability test.
+ENV_VARIABLES = {
+    "portability_service_url": os.environ.get("PORTABILITY_SERVICE_URL"),
+    "portability_service_headers": os.environ.get("PORTABILITY_SERVICE_HEADERS")
+}
+
+def verify_environment_variables():
+    """Check that required environment variables are defined."""
+
+    errors = []
+
+    for key, value in ENV_VARIABLES.items():
+        if not value:
+            errors.append("Environment variable {} is undefined".format(key.upper()))
+    if errors:
+        for error in errors:
+            sys.stderr.write(error + "\n")
+        raise TestEnvironmentError(errors)
+
+def submit_portability_test(test_inputs):
+    """Submit the portability to the service and return the test's id."""
+
+    service_headers = json.loads(ENV_VARIABLES["portability_service_headers"])
+    test_endpoint = ENV_VARIABLES["portability_service_url"] + "/portability_tests"
+    response = requests.post(
+        test_endpoint,
+        headers=service_headers,
+        json=test_inputs)
+
+    print("Portability service response:\n{}".format(
+        json.dumps(json.loads(response.text), indent=4)), flush=True)
+
+    test_id = json.loads(response.text)["test_id"]
+
+    return test_id
+
+def print_test_states(test_states):
+    """Print the states of tests."""
+    print("{:%Y-%m-%dT%H:%M:%SZ}".format(datetime.datetime.utcnow()), flush=True)
+    for test_name, test_state in test_states.items():
+        print("TEST:{} STATE:{}".format(test_name, test_state), flush=True)
+
+def monitor_tests(test_ids):
+    """Check the status of tests until they all either fail or succeed. If any fail, raise
+    TestFailure, and if they all succeed, return successfully.
+    """
+
+    service_headers = json.loads(ENV_VARIABLES["portability_service_headers"])
+
+    terminal_tests = {}
+    while True:
+        time.sleep(120)
+
+        test_states = {}
+
+        for test_name, test_id in test_ids.items():
+
+            if test_name in terminal_tests:
+                test_states[test_name] = terminal_tests[test_name]
+                continue
+
+            status_endpoint = ENV_VARIABLES["portability_service_url"]  + "/portability_tests/" + \
+                test_id + "/status"
+
+            response = requests.get(status_endpoint, headers=service_headers)
+            test_state = json.loads(response.text)["state"]
+            test_states[test_name] = test_state
+
+        print_test_states(test_states)
+
+        for test_name, test_state in test_states.items():
+            if test_state in ("Failed", "Succeeded"):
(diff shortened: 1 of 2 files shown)


smartseq2_single_sample/ss2_checker.wdl
```diff
@@ -0,0 +1,15 @@
+task SmartSeq2Checker {
+  File rna_metrics
+
+  String expected_rna_metrics_hash
+
+  command <<<
+    rna_metrics_hash=$(tail -n +6 ${rna_metrics} | md5sum | awk '{print $1}')
+    
+    if [ "$rna_metrics_hash" != "${expected_rna_metrics_hash}" ]; then
```
jishuxu: should it be `"${rna_metrics_hash}"` instead of `"$rna_metrics_hash"` ?

smartseq2_single_sample/ss2_checker.wdl
```diff
@@ -0,0 +1,15 @@
+task SmartSeq2Checker {
+  File rna_metrics
+
+  String expected_rna_metrics_hash
+
+  command <<<
+    rna_metrics_hash=$(tail -n +6 ${rna_metrics} | md5sum | awk '{print $1}')
```
jishuxu: you might add `sed -e ` before multiple bash script lines.

smartseq2_single_sample/ss2_checker.wdl
mckinsel: I think it would be the same, but I prefer to use different syntax for bash variables and WDL variables when I can. However, has everyone else settled on using braces around bash variables?

smartseq2_single_sample/ss2_checker.wdl
mckinsel: Thanks, I'll add it.

mckinsel:
Hey @jishuxu, thanks for taking a look. Here are some answers:

1. I don't think any environment other than Cromwell+Google can interpret `preemptible`. The default behavior for other environments it to just ignore fields they can't use. Part of the purpose of these tests is to expose these sorts of couplings. Also, I should mention that the infrastructure for launching these tests is within AWS, but the environments themselves could be anything. There's currently one environment implemented, and that's Cromwell with a local backend. That's running on an AWS instance, but it doesn't know anything about AWS.

2. I found the hash of the `rna_metrics` output from the SS2 run you sent me. I provide that as in input to the portability workflow. The checker tool verifies that hash of the `rna_metrics` file produced in external environments matches that expected hash. It's working with the local Cromwell environment.

smartseq2_single_sample/ss2_checker.wdl
jishuxu: Yeah, it should be the same. To add `{}` just just to make variables in wdl consistent, pure cosmetics effect.

jishuxu:
Good to know the `rna_metrics` are consistent between two runs. Thanks!

dshiga:
Thanks @mckinsel. Can you also link to the elastc ticket if there is one for this?

smartseq2_single_sample/ss2_portability.wdl
```diff
@@ -0,0 +1,48 @@
...
+      sample_name=sample_name,
+      output_name=output_name,
+      fastq1=fastq1,
+      fastq2=fastq2
+  }
+
+  call ss2_checker_wdl.SmartSeq2Checker as ss2_checker {
+    input:
+      rna_metrics=ss2_target.rna_metrics,
+      expected_rna_metrics_hash=expected_rna_metrics_hash
```
samanehsan: Why does this only check the `rna_metrics` and not the other outputs of the SmartSeq2 pipeline?

smartseq2_single_sample/ss2_portability_inputs.json
```diff
@@ -0,0 +1,17 @@
+{
+  "SS2PortabilityWorkflow.gtf_file": "https://s3.amazonaws.com/mk-hca-portability-data/new_ss2/gencode.v27.primary_assembly.annotation.gtf",
```
samanehsan: Should we use the input files that are in the mint team google cloud storage instead?

samanehsan:
What do you think of creating a `test/smartseq2_single_sample/portability` folder and adding in the ss2_portability.wdl and all of its relevant input files there? Like how the optimus tests have folders for scientific and PR testing: https://github.com/HumanCellAtlas/skylab/tree/master/test/optimus

samanehsan:
Just a few comments/questions otherwise this looks great @mckinsel!

samanehsan:
Oh there's one more comment I forgot to add! Could you please update the readme to include a note about the portability tests and how to run them?

smartseq2_single_sample/ss2_portability_inputs.json
TimothyTickle: Longterm storage I would suggest using the Prod Blue box, until then I imagine solutions are temporary.

mckinsel: Yeah, we need a publicly-accessible place to store the test data. Right now, I believe all the Mint google buckets are private, so they won't work. The prod blue box makes sense (though how do we ingest that?). In the meantime, it would be great to have a public place in google cloud storage where we could put all of our test data.

smartseq2_single_sample/ss2_checker.wdl
```diff
@@ -0,0 +1,16 @@
+task SmartSeq2Checker {
+  File rna_metrics
+
+  String expected_rna_metrics_hash
+
+  command <<<
+    set -e
+    rna_metrics_hash=$(tail -n +6 ${rna_metrics} | md5sum | awk '{print $1}')
```
TimothyTickle: Is this only checking the tail of the metrics file? and if so, is there a reason the full file is not checked?

mckinsel: It's checking all but the first six lines. That's what the `+` does in `-n +6`. The first six lines are paths and timestamps which we don't expect to be the same across executions.

smartseq2_single_sample/ss2_portability.wdl
mckinsel: This is mostly just a demo of a test. Once #69 is merged, I plan to just run that.

mckinsel:
Thanks @samanehsan, I'll move the test into `test/smartseq2_single_sample/portability` and update the README.

ambrosejcarr:
When you rebase this against master, the failing SS2 test should succeed.

rexwangcc:
LGTM and test pull approve

dshiga:
Looks like it needs rebasing to fix failing jenkins tests. The ss2 test failed with:
```
IOError: [Errno 2] No such file or directory: '/working/test/smartseq2_single_sample/pr/test_smartseq2_single_sample_PR.wdl
```

mckinsel:
jenkins retest

test/run_portability_test.py
```diff
@@ -0,0 +1,199 @@
...
+    for key, value in dependencies_dict.items():
+        try:
+            loaded_dependencies.append({
+                "name": key,
+                "code": open(value).read()
+            })
+        except IOError:
+            errors.append("Count not read dependency {}".format(value))
+
+    if errors:
```
ambrosejcarr: I like this -- waiting to throw errors until you need to could produce a nice productivity boost for the user.

test/run_portability_test.py
```diff
@@ -0,0 +1,199 @@
...
+
+# This is what we need to find in a test directory. I guess we'll glob the PR
+# test WDL itself since it's named after the pipeline.
+TEST_DIR_LAYOUT = {
+    "inputs": "test_inputs.json",
+    "dependencies": "dependencies.json",
+    "test": "*PR.wdl"
+}
+
+# These are the environment variables we're expecting to be able to submit a
```
ambrosejcarr: I'm of two minds here: you normally define these things at the tops of files, but you don't use them until after you're done testing the inputs, and I had to scroll back up to remind myself about what they are. If it were me I'd be tempted to move them down for readability.

test/run_portability_test.py
```diff
@@ -0,0 +1,199 @@
...
+    print("Portability service response:\n{}".format(
+        json.dumps(json.loads(response.text), indent=4)), flush=True)
+
+    test_id = json.loads(response.text)["test_id"]
+
+    return test_id
+
+def print_test_states(test_states):
+    """Print the states of tests."""
+    print("{:%Y-%m-%dT%H:%M:%SZ}".format(datetime.datetime.utcnow()), flush=True)
```
ambrosejcarr: Did you find that you needed to use `flush=True` to get it to properly write to stdout? I've never found I needed to use this.

test/run_portability_test.py
```diff
@@ -0,0 +1,199 @@
...
+                continue
+
+            status_endpoint = ENV_VARIABLES["portability_service_url"]  + "/portability_tests/" + \
+                test_id + "/status"
+
+            response = requests.get(status_endpoint, headers=service_headers)
+            test_state = json.loads(response.text)["state"]
+            test_states[test_id] = test_state
+
+        print_test_states(test_states)
```
ambrosejcarr: Is it possible to log the failures the same way you do for inputs? I'm wondering about a case where you have more than one failing test and want to be explicit about _all_ the failing tests. I see that the print information will contain this but I wonder if the user will notice over the `TestFailure()`.

test/run_portability_test.py
mckinsel: I had to do that for travis or else it would kill the job after 10 minutes of what looked like inactivity.

ambrosejcarr: Interesting. How does flush produce "activity"? Something weird about how travis interacts with the stdout stream? (not important, just interested)

test/run_portability_test.py
mckinsel: I'm pretty sure that's how it does work. The `TestFailure` mostly just gives the whole script a non-zero exit code if any of the tests fail. That way you get a failure reported via CircleCI.

But, it should run each test to completion and report the status of each one. So if you do see an overall failure, you can look in the CI log and see what test(s) failed. It looks like this:

```
TEST:test/smartseq2_single_sample/pr/ STATE:Succeeded
TEST:test/optimus/pr STATE:Failed
```

test/run_portability_test.py
mckinsel: Yeah, travis might be interacting with stdout differently. This is the test that was fixed by adding `flush=True`: https://travis-ci.org/HumanCellAtlas/skylab/builds/326686167.

Adding flush changed it to this: https://travis-ci.org/HumanCellAtlas/skylab/builds/326688523

test/run_portability_test.py
mckinsel: Do you have a preferred place to put it? References to them are a little bit scattered throughout the code.

ambrosejcarr: I was thinking line 104.

test/run_portability_test.py
ambrosejcarr: Weird! ok, thanks for showing.

test/run_portability_test.py
ambrosejcarr: Ah, got it. Great.

test/run_portability_test.py
```diff
@@ -0,0 +1,199 @@
...
+    dependencies_dict = json.loads(dependencies_json_string)
+
+    for key, value in dependencies_dict.items():
+        try:
+            loaded_dependencies.append({
+                "name": key,
+                "code": open(value).read()
+            })
+        except IOError:
+            errors.append("Count not read dependency {}".format(value))
```
dshiga: You probably meant "could" instead of "count" here, right?

mckinsel: Oh yeah, good catch.

test/run_portability_test.py
mckinsel: Okay thanks, I'll move it down.

PR #69, opened 2018-01-16
added ss2 PR test

ambrosejcarr:
This PR contains code to run a test against the SS2 pipeline on PR. It is deterministic and generates a successful result when run in cromwell. 

The notebook will be removed before the branch gets merged; it is just here for testing purposes. 

It is not yet implemented in jenkins, pending some assistance from devops.

### DIFF
--- a/test/smartseq2_single_sample/pr/TestSmartSeq2SingleCellPR.wdl
+++ b/test/smartseq2_single_sample/pr/TestSmartSeq2SingleCellPR.wdl
@@ -1,0 +1,53 @@
+import "ss2_single_sample.wdl" as target_wdl
+import "ValidateSmartSeq2SingleCell.wdl" as checker_wdl
+
+# this task will be run by the jenkins script that gets executed on our PRs.
+workflow TestSmartSeq2SingleCellPR {
+
+  # expected hashes of target_workflow outputs
+  String expected_counts_hash
+  String expected_alignment_summary_metrics_hash
+
+  # SS2 inputs
+  File gtf_file
+  File genome_ref_fasta
+  File rrna_intervals
+  File gene_ref_flat
+  File hisat2_ref_index
+  File hisat2_ref_trans_index
+  File rsem_ref_index
+  File hisat2_ref_name
+  File hisat2_ref_trans_name
+  String stranded
+  String sample_name
+  String output_name
+  File fastq1
+  File fastq2
+
+  call target_wdl.SmartSeq2SingleCell as target_workflow {
+    input:
+      gtf_file = gtf_file,
+      genome_ref_fasta = genome_ref_fasta,
+      rrna_intervals = rrna_intervals,
+      gene_ref_flat = gene_ref_flat,
+      hisat2_ref_index = hisat2_ref_index,
+      hisat2_ref_trans_index = hisat2_ref_trans_index,
+      rsem_ref_index = rsem_ref_index,
+      hisat2_ref_name = hisat2_ref_name,
+      hisat2_ref_trans_name = hisat2_ref_trans_name,
+      stranded = stranded,
+      sample_name = sample_name,
+      output_name = output_name,
+      fastq1 = fastq1,
+      fastq2 = fastq2
+  }
+
+  call checker_wdl.ValidateSmartSeq2SingleCell as checker_workflow {
+    input:
+     counts = target_workflow.rsem_gene_results,
+     expected_counts_hash = expected_counts_hash,
+     alignment_summary_metrics = target_workflow.alignment_summary_metrics,
+     expected_alignment_summary_metrics_hash = expected_alignment_summary_metrics_hash
+  }
+
+}
--- a/test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl
+++ b/test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl
@@ -1,0 +1,41 @@
+task ValidateSmartSeq2SingleCell {
+      File counts
+      String expected_counts_hash
+
+      File alignment_summary_metrics
+      String expected_alignment_summary_metrics_hash
+
+  command <<<
+
+    # catch intermittent failures
+    set -eo pipefail
+
+    # calculate hashes; awk is used to extract the hash from the md5sum output that contains both
+    # a hash and the filename that was passed. We parse the first 7 columns because a bug in RSEM
+    # makes later columns non-deterministic.
+    counts_hash=$(cut -f 1-7 "${counts}" | md5sum | awk '{print $1}')
+
+    # this parses the picard metrics file with awk to remove all the run-specific comment lines (#)
+    picard_metrics_hash=$(cat "${alignment_summary_metrics}" | awk 'NF && $1!~/^#/' | md5sum | awk '{print $1}')
+
+    if [ "$counts_hash" != "${expected_counts_hash}" ]; then
+      >&2 echo "counts_hash ($counts_hash) did not match expected hash (${expected_counts_hash})"
+      fail=true
+    fi
+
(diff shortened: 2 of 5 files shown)


test/test_pr_tests.ipynb
```diff
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 14,
+   "execution_count": 1,
```
rexwangcc: This notebook looks good to me. The only suggestion I could give here is to add workflow labels(such as `"labels":"PR-test"`) to those testing workflows by using the PATCH endpoint, to distinguish them from other workflows.

ambrosejcarr: Got it. I will probably remove the notebook before I merge, as the workflow will be executed with Jenkins. Because I have a few computers that I develop on, I tend to keep the testing code paired with the branches I work on. 

However, I think the repo is cleaner without the notebooks, so will usually remove them unless they contain analysis products (i.e. I'm working on a research ticket) . 

I've added a note to the PR; forgot this time. Sorry!

test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl
```diff
@@ -0,0 +1,43 @@
...
+
+    if [ $fail == "true" ]; then exit 1; fi
+
+  >>>
+  
+  runtime {
+    docker: "ubuntu:zesty"
+    cpu: 1
+    memory: "3.75 GB"
+    disks: "local-disk 2 HDD"
```
jsotobroad: as a random piece of information, with Google you cant ask for less than 10 GB (this wont fail it will just be defaulted to 10)

test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl
```diff
@@ -0,0 +1,43 @@
...
+    # a hash and the filename that was passed. We parse the first 7 columns because a bug in RSEM
+    # makes later columns non-deterministic.
+    counts_hash=$(cut -f 1-7 "${counts}" | md5sum | awk '{print $1}')
+
+    # the picard metrics are contained in a .tar.gz file; in addition to the above processing,
+    # this unzips the archive, and parses it with awk to remove all the run-specific comment lines (#)
+    picard_metrics_hash=$(cat "${alignment_summary_metrics}" | awk 'NF && $1!~/^#/' | md5sum | awk '{print $1}')
+
+    if [ "$counts_hash" != "${expected_counts_hash}" ]; then
+      >&2 echo "counts_hash ($counts_hash) did not match expected hash (${expected_counts_hash})"
```
jsotobroad: I'm going back and forth as to whether I would want to see the diff here or not along with the mismatching hashes just to provide a little more debugging information or do we expect someone to dig into the files themselves if their test fails?

test/smartseq2_single_sample/pr/TestSmartSeq2SingleCellPR.wdl
```diff
@@ -0,0 +1,59 @@
...
+  File hisat2_ref_name
+  File hisat2_ref_trans_name
+  # samples
+  String stranded
+  String sample_name
+  String output_name
+  File fastq1
+  File fastq2
+
+  call target.SmartSeq2SingleCell as target {
```
jsotobroad: I would be a little weary of aliasing your call with the same alias as your import if nothing else just for readability

test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl
```diff
@@ -0,0 +1,43 @@
+
```
jsotobroad: extra `\n`

test/test_pr_tests.ipynb
jsotobroad: i need to learn more about this python notebook stuff :(

test/smartseq2_single_sample/pr/TestSmartSeq2SingleCellPR.wdl
```diff
@@ -0,0 +1,59 @@
+import "ss2_single_sample.wdl" as target
+import "ValidateSmartSeq2SingleCell.wdl" as checker
+
+
```
jsotobroad: extra `\n` (unless done on purpose)

test/smartseq2_single_sample/pr/TestSmartSeq2SingleCellPR.wdl
ambrosejcarr: Interesting, we do this a lot. How would you fix this? Would `target_wdl` for the import and `target_workflow` for the execution be a good solution?

test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl
ambrosejcarr: Well, one concern is that these files could sometimes be binary, in what case a diff would overload the log. What about the head of a diff? 

I do imagine that people would often need to dig into their files, but wouldn't be opposed to making life easier in cases where a diff solves the problem.

test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl
ambrosejcarr: fixed.

ambrosejcarr:
jenkins retest

ambrosejcarr:
@jsotobroad I thought a bit more about the `diff` request. I think it could be difficult to implement since we run against hashes, and don't actively run the `master` pipelines. Thus, we would also need to pass "successful" outputs, which would increase complexity, and I'm not sure where we'd store those files. 

Happy to converse more about this. If you're satisfied with this response, then I think I've addressed all your comments (thanks!).

jsotobroad:
That makes a lot of sense to me, thanks for clearing the diff stuff, I aggre.  I really like the `wdl` / `workflow` distinctions in your aliasing 👍  👍

test/smartseq2_single_sample/pr/dependencies.json
```diff
@@ -0,0 +1,9 @@
+{
+  "ss2_single_sample.wdl": "/working/smartseq2_single_sample/ss2_single_sample.wdl",
+  "ValidateSmartSeq2SingleCell.wdl": "/working/test/smartseq2_single_sample/pr/ValidateSmartSeq2SingleCell.wdl",
+  "hisat2_QC_pipeline.wdl": "/working/smartseq2_single_sample/pipelines/hisat2_QC_pipeline.wdl",
+  "hisat2_rsem_pipeline.wdl": "/working/smartseq2_single_sample/pipelines/hisat2_rsem_pipeline.wdl",
+  "hisat2.wdl": "/working/pipelines/tasks/hisat2.wdl",
+  "picard.wdl": "/working/pipelines/tasks/picard.wdl",
+  "rsem.wdl": "/working/pipelines/tasks/rsem.wdl"
+}
```
mckinsel: What's going on with these paths? Does the repo get cloned into `/working` when you run the tests in jenkins?

ambrosejcarr: Correct.

ambrosejcarr: Working is also set as `cwd`.

mckinsel: Thanks for clarifying. I think this is probably letting the details of how jenkins is executing the test bleed into the definition of the test. But we can address it when we reorganize the repo.

ambrosejcarr: Sounds good...  I think the long term goal is to run the portability service in lieu of these tests, but would be interested to hear your thoughts during the reorg discussion. :+1:

PR #73, opened 2018-01-19
initial reorg

ambrosejcarr:
Let's discuss how we want to move files around.


ambrosejcarr:
I believe we came to some agreements, please correct me if any of these are wrong: 
1. Analysis products (notebooks, images) are to be moved to a new repository `skylab-analysis`
2. The existing top-level pipelines folders (optimus, smartseq2_single_sample) will move to a `pipelines` folder, in which they will be subfolders
3. The existing `pipelines` folder will become `library` and will house `tasks`, `subworkflows` and `accessories`
4. Benchmarking pipelines and tools will go into `accessories`
5. Monitoring scripts that were in the top level will go into `cromwell-tools`. 
6. We will make use of `dependencies.json` files for any top-level pipeline that is intended to be executed in order to make dependencies explicit (versus relative)

samanehsan:
We'll need to update some of the file paths defined in the lira config here as well: https://github.com/HumanCellAtlas/secondary-analysis/blob/master/test/create_lira_config.py#L37

kbergin:
@ambrosejcarr There is a duplicate of /accessories in pipelines and in library it seems

kbergin:
I think the `smartseq2_single_sample/analysis` dir that is currently at the top level should be something else. I think everything should fit into `library` `pipelines` or `test` or should be moved to skylab-analysis. Thoughts?

Should the files in `src/calculate_cost_workflow.py` and `src/parse_picard.py` be in other repos? Maybe pipeline tools or cromwell tools?

ambrosejcarr:
@kbergin Agree. I'll make another stab at the reorg in a few minutes and see if I can get the PR tests running against the files here.

kbergin:
We also still have duplication in accessory-workflows and pipelines/accessories

ambrosejcarr:
@kbergin I am removing the files in src that were added in @jishuxu recent commit. Those files are SS2 specific subworkflow cost objects, and I'm not 100% certain they will run on the updated SS2 workflow (though they might, I couldn't make them work on optimus). We need a general solution wherein the cost script follows arbitrary subworkflows. When fixed, I concur that they could go into analysis.

jishuxu:
Good to go

ambrosejcarr:
@samanehsan @jishuxu You guys have both said this is good to merge, could I get a 👍 ? 

After that I'll merge it in. Thanks!

PR #74, opened 2018-01-26
Fix dockerfiles (518)

rexwangcc:
Fix dockerfiles after Ubuntu zesty reached its end of life cycle.

Specifically, for samtools, refactor the context of the dockerfile to use samtools version 1.6

### DIFF
--- a/docker/analysis-tools/Dockerfile
+++ b/docker/analysis-tools/Dockerfile
@@ -71,10 +71,10 @@ RUN Rscript -e "install.packages('ggplot2')"
 RUN Rscript -e "install.packages('googleCloudStorageR')"
 
 # Mount Google Cloud secret key
-ADD broad-dsde-mint-dev-testR.json /usr/secrets/broad-dsde-mint-dev-testR.json
+ADD broad-dsde-mint-dev-analysis-tools.json /usr/secrets/broad-dsde-mint-dev-analysis-tools.json
 
 # Set the environment variable for Python
-ENV GOOGLE_APPLICATION_CREDENTIALS=/usr/secrets/broad-dsde-mint-dev-testR.json
+ENV GOOGLE_APPLICATION_CREDENTIALS=/usr/secrets/broad-dsde-mint-dev-analysis-tools.json
 
 # Set the environment variable for R
 ENV GCS_AUTH_FILE=$GOOGLE_APPLICATION_CREDENTIALS
@@ -85,3 +85,13 @@ RUN curl -sSL https://sdk.cloud.google.com > /tmp/gcl && bash /tmp/gcl --install
 # Configure gcloud
 ENV PATH $PATH:/usr/gcloud/google-cloud-sdk/bin
 RUN gcloud auth activate-service-account --key-file=$GCS_AUTH_FILE
+
+# Add Cromwell Credentials
+ADD broad-dsde-mint-dev-cromwell.json /usr/secrets/broad-dsde-mint-dev-cromwell.json
+
+RUN wget https://ftp-trace.ncbi.nlm.nih.gov/sra/sdk/2.8.2-1/sratoolkit.2.8.2-1-ubuntu64.tar.gz \
+	&& tar -zxvf sratoolkit.2.8.2-1-ubuntu64.tar.gz \
+	&& cp -r  sratoolkit.2.8.2-1-ubuntu64/ /usr/local/
+ENV PATH /usr/local/sratoolkit.2.8.2-1-ubuntu64/bin:$PATH
+
+RUN pip3 install ipython
--- a/docker/kallisto/Dockerfile
+++ b/docker/kallisto/Dockerfile
@@ -11,7 +11,13 @@ RUN apt update && apt install -y \
   cmake \
   curl \
   libhdf5-dev \
-  samtools
+  libncurses5-dev \
+  openssl \
+  liblzma-dev \
+  libbz2-dev \
+  bzip2 \
+  libcurl4-openssl-dev \
+  libssl-dev
 
 RUN mv /usr/include/hdf5/serial/* /usr/include/
 
@@ -28,3 +34,14 @@ RUN curl -L -o kallisto-0.43.1.tar.gz \
   cd / && \
   rm -r kallisto-0.43.1 && \
   rm kallisto-0.43.1.tar.gz
+
+# Install samtools
+WORKDIR /usr/local/samtools
+ADD https://github.com/samtools/samtools/releases/download/1.6/samtools-1.6.tar.bz2 .
+
+RUN tar -xvf samtools-1.6.tar.bz2 && \
+    rm samtools-1.6.tar.bz2 && \
+    cd samtools-1.6 && \
+    ./configure --prefix=/usr && \
+    make && \
+    make install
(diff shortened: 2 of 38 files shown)


docker/samtools/Dockerfile
```diff
@@ -1,10 +1,32 @@
-FROM ubuntu:zesty
+FROM ubuntu:16.04
```
rexwangcc: @ambrosejcarr I want to let you look into this dockerfile change specifically.

docker/patched-star/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ubuntu:zesty
+FROM ubuntu:16.04
```
rexwangcc: @ambrosejcarr This is also a risky change. 

We use `apt install samtools` in this dockerfile, without specifying the version of the samtools, this is dangerous. 

Ubuntu 17.04 provides: samtools Version: 1.3.1 (using htslib 1.3.2) BUT 
Ubuntu 16.04 only provides: samtools Version: 0.1.19-96b5f2294a

we should aware this and think about if there are any scientific packages in other images that are installed by `apt install` but with no version specified. If you don't mind, I'm going to modify this dockerfile to use samtools 1.6, build it from source, like what I did in the samtools dockerfile.

jishuxu: just make sure the syntax of samtools are consistent between these two version.

docker/picard/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM java:openjdk-8-jre
+FROM openjdk:8-jre
```
dshiga: Can you explain the reason for this change?

docker/patched-star/Dockerfile
```diff
@@ -7,8 +7,17 @@ LABEL maintainer="Ambrose J. Carr <acarr@broadinstitute.org>" \
   website="https://github.com/alexdobin/STAR"
 
 RUN apt update && apt install -y \
+  g++ \
```
dshiga: I assume these are all dependencies of samtools. If that's the case, can you install star before we install these dependencies, just so it's clear they're for samtools?

docker/kallisto/Dockerfile
```diff
@@ -11,7 +11,13 @@ RUN apt update && apt install -y \
   cmake \
   curl \
   libhdf5-dev \
-  samtools
+  libncurses5-dev \
```
dshiga: Is there a way to indicate which tool these are dependencies of? Otherwise people will probably keep them in here forever, even if, for example, we remove samtools later, because people won't be sure what they're for.

docker/patched-star/Dockerfile
rexwangcc: Yes, most of them are dependencies of samtools 1.6, I will re-organize the order to make it clearer for this docker and other necessary dockerfiles. Thanks for this suggestion!

docker/analysis-tools/Dockerfile
```diff
@@ -85,3 +85,13 @@ RUN curl -sSL https://sdk.cloud.google.com > /tmp/gcl && bash /tmp/gcl --install
 # Configure gcloud
 ENV PATH $PATH:/usr/gcloud/google-cloud-sdk/bin
 RUN gcloud auth activate-service-account --key-file=$GCS_AUTH_FILE
+
+# Add Cromwell Credentials
+ADD broad-dsde-mint-dev-cromwell.json /usr/secrets/broad-dsde-mint-dev-cromwell.json
```
dshiga: I thought you were just updating to switch to 16.04 -- why is this being added here? Similarly for sratoolkit below.

docker/kallisto/Dockerfile
rexwangcc: Without looking at other mature big dockers, I come up with two ideas to address this:
1. Elaborate comments in dockerfiles.
2. Use separate `apt install` command for each tool and its dependencies. (like one command block  for star and another for samtools)
But I will check if there are any other better solutions used in other big dockers.

dshiga:
Thanks for stepping up to fix this so promptly @rexwangcc!

docker/analysis-tools/Dockerfile
rexwangcc: Actually, the analysis-tools in master branch is outdated, @jishuxu updated this docker in her branch, and I just want to include those changes here together with the Ubuntu version update.

docker/picard/Dockerfile
rexwangcc: Sure, according to https://hub.docker.com/_/java/ , the `java` base image is officially deprecated in favor of the `openjdk` image

rexwangcc:
@dshiga Thanks for all the helpful comments and the prompt review! 

> Do you want to change the wdls to use the new docker images so the PR tests will run and catch potential problems? Not sure how much work that is. But not sure how to be sure these changes don't break anything without either that or manual testing.

This is a really great question. I have been thinking about this problem as well while updating the dockers. We have two pains here (and also in the future):
1. We need to manually change the docker tags in all the wdls after docker tags are changed/increased. This task will evolve even more manual work if we have more and more wdls in the future.
2. We still don't have a good way to test if only the docker images are changed. We can do it manually, but again it is not efficient.

I made a follow-up bookmark ticket (#541) for this ticket to update wdls manually. But I think it would be really great if we could address the above two problems with automation. For example, we can have a script for us to update/replace docker tags in wdls, and we could have some incremental/partial Integration Testing to only test docker changes.

docker/patched-star/Dockerfile
jishuxu: This docker or STAR  probably is also obsoleted. Ambrose should give a look later if we can build STAR docker from latest release.

docker/analysis-tools/Dockerfile
jishuxu: it is analysis docker, not stored in quay. I added more packages as I needed to run analysis.

mckinsel:
Hey I'm not too familiar with how these images get built by quay, but do we have an easy way to run the SS2 test with the new images before merging?

rexwangcc:
@mckinsel I believe if the Jenkins integration test and the SS2 PR test pass, then the wdls and new images should be good to merge.

rexwangcc:
All the tests passed now, I'm going to merge the changes into master. In case that there are corner cases that the tests don't cover, I list the current docker tracking table below, all wdls are using docker images shown as:

- RSEM: `quay.io/humancellatlas/secondary-analysis-rsem:v0.2.2-1.3.0`
- Samtools: `quay.io/humancellatlas/secondary-analysis-samtools:v0.2.2-1.6`
- Star: `quay.io/humancellatlas/secondary-analysis-star:v0.2.2-2.5.3a-40ead6e`
- Patched-star: `quay.io/humancellatlas/secondary-analysis-star:v0.2.2-2.5.3a-1.0.0`
- Kallisto: `quay.io/humancellatlas/secondary-analysis-kallisto:v0.2.2-0.43.1`
- Dropseqtools: `quay.io/humancellatlas/secondary-analysis-dropseqtools:v0.2.2-1.12`
- Picard: `quay.io/humancellatlas/secondary-analysis-picard:v0.2.2-2.10.10`
- Subread: `quay.io/humancellatlas/secondary-analysis-subread:v0.2.2-1.6.0`
- Hisat2: `quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0`
- Python3-scientific: `quay.io/humancellatlas/secondary-analysis-python3-scientific:0.1.5`

PR #77, opened 2018-02-02
[Do Not Merge] First stab at a Mint wdl style'ish guide...ish

jsotobroad:
I wanted to try to take a stab at what kind of style guide the team might want to follow for our wdls.
I've put most of my suggestions in `SS2SingleSample.wdl` but you can find anything I've added by looking for the pattern`## **`.  I've stated some things and have also asked a few questions about things I am not sure about but comments on anything would be really awesome and I fully expect for whats written here to change quite a bit.

I also took the styles I wrote up (totally biased on how I like to write my wdls) and applied them to the SS2 workflow as a whole in this PR as well so at least we could get an idea of what they look like.

I don't plan on merging this in ever but it might be a good first step in figuring out what we want to make our WDLs look like and then we can go back and apply those guidelines to already written WDLs and future WDLs.

This will also be good feedback for the WDL team as I know they are looking to work on a wdl style/possibly linter in the near future.


pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+import "Hisat2RsemPipeline.wdl" as RunHisatRSEMWDL
+
+## ** Explanations above the workflow about what inputs it expects, what outputs/metrics it outputs
+## ** What, if any, assumptions are made in the pipeline
+## ** In general any kind of description you want to add above the workflow is nice
+## ** A good example of this can be found in these wdls
+## ** https://github.com/broadinstitute/dsde-pipelines/blob/develop/genomes_in_the_cloud/single_sample/PairedSingleSampleWf.wdl
+## ** https://github.com/broadinstitute/gatk/blob/master/scripts/mutect2_wdl/mutect2.wdl
+
+## ** Should comments / descriptions above a task definition be mandatory?
```
kbergin: I vote yes!

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+## ** This will change quite a bit i believe when the new spec of WDL comes out (WDL SPEC 3) but we should define all values we care about
+## ** even if they are the default values.  e.g. write `cpu: 1` instead of leaving it out because its the default.
+
+##### Comment Rules ######
+## **  Should comments have one or two `##` before a comment.  Maybe two for workflow or task descriptions and one for any other kind of comment?
+
+import "Hisat2QcPipeline.wdl" as RunHisatQCWDL  ## ** Should import alias follow some kind of pattern like ending in "WDL" or something else
+import "Hisat2RsemPipeline.wdl" as RunHisatRSEMWDL
+
+## ** Explanations above the workflow about what inputs it expects, what outputs/metrics it outputs
```
kbergin: +1

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+## ** What, if any, assumptions are made in the pipeline
+## ** In general any kind of description you want to add above the workflow is nice
+## ** A good example of this can be found in these wdls
+## ** https://github.com/broadinstitute/dsde-pipelines/blob/develop/genomes_in_the_cloud/single_sample/PairedSingleSampleWf.wdl
+## ** https://github.com/broadinstitute/gatk/blob/master/scripts/mutect2_wdl/mutect2.wdl
+
```
kbergin: I think in this top comment block we'd also want to have information about licensing, workflow structure, author

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
cjllanwarne: This would be better included in `meta` and `parameter_meta` blocks

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+## ** Space before and after operators e.g. `+`, `=`
+## ** Space after `:` character in runtime block
+## ** Space after `#` when creating a comment (or `##` if we decide thats our comment definer)
+
+##### Tab vs Space for Indents ######
+## ** Always use spaces? I think whats most frequently used is whenever you want to indent you use 2 spaces but we should decide what works best for us.
+
+##### Runtime Attributes ######
+## ** This will change quite a bit i believe when the new spec of WDL comes out (WDL SPEC 3) but we should define all values we care about
+## ** even if they are the default values.  e.g. write `cpu: 1` instead of leaving it out because its the default.
```
cjllanwarne: This is subtly different from leaving it unsaid.

Leaving it unsaid allows a runner to override it whereas I believe specifying it hard-codes it into the WDL.

For things like `failOnStderr` that *should* be baked into the workflow whereas for things like `cpu` it should be up to whoever's running it.

To allow overrides you can have runtime fed by inputs with defaults, eg
```
task foo {
  Int cpus
  runtime {
    cpu: cpus
  }
}
```

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
cjllanwarne: This sort of information would be better included in meta and parameter_meta blocks

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
cjllanwarne: In parameter meta, it would be possible in the future to add an option to womtool to validate that every input has a description, etc. Leaving that in comments precludes validation.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: But if its a single threaded tool why would you want to allow the runner to override it?  I think theres definitely cases where you want it to be overridable but my original comment was more so talking about not relying on defaults set by Cromwell, i.e. if you don't set a `cpu` value for the PAPI backend you always get a value of `1`.  I don't even know the defaults for local or any other backend which is kind of my point because if we had set a value in our task I woudln't have to worry about that either.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: I always forget about the meta/parameter_meta blocks!

jsotobroad:
@LeeTL1220 @yfarjoun your input would be appreciated as well.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
+##### Naming Rules #####
+## ** Files should be named in CamelCase starting with a capital
```
yfarjoun: What are the capitalization rules when using an acronym (like SS, WGS, VCF, etc.)?

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+## ** Always use spaces? I think whats most frequently used is whenever you want to indent you use 2 spaces but we should decide what works best for us.
+
+##### Runtime Attributes ######
+## ** This will change quite a bit i believe when the new spec of WDL comes out (WDL SPEC 3) but we should define all values we care about
+## ** even if they are the default values.  e.g. write `cpu: 1` instead of leaving it out because its the default.
+
+##### Comment Rules ######
+## **  Should comments have one or two `##` before a comment.  Maybe two for workflow or task descriptions and one for any other kind of comment?
+
+import "Hisat2QcPipeline.wdl" as RunHisatQCWDL  ## ** Should import alias follow some kind of pattern like ending in "WDL" or something else
```
samanehsan: I think adding `WDL` to the import alias seems redundant

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
samanehsan: 👍

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
kbergin: To play devil's advocate, I think one benefit of saying `RunHisatQCWDL` instead of `RunHisatQC` is that below in line 71 it's clear that a subworkflow is being called vs a task. What are your thoughts? 

I think I'd choose `RunHisatQCWDL.RunHisat2QCPipeline as SmartSeq2QCTrack` over how it is below. I'm probably not capitalizing correctly in my above example.

library/tasks/Hisat2.wdl
```diff
@@ -46,35 +46,38 @@ task HISAT2PE {
...
     samtools sort -@ 4 -O bam -o "${output_name}.bam" "${output_name}.sam" 
   }
+
   runtime {
-    docker:"quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0"
-    memory:"5 GB"
+    docker: "quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0"
+    memory: "5 GB"
     disks: "local-disk " + disk_size + " HDD"
     cpu: "4"
```
ambrosejcarr: I believe cromwell accepts simple integers here instead of strings. I feel integers are clearer for this field.

library/tasks/Hisat2.wdl
```diff
@@ -46,35 +46,38 @@ task HISAT2PE {
       -p 4 -S ${output_name}.sam 
     samtools sort -@ 4 -O bam -o "${output_name}.bam" "${output_name}.sam" 
   }
+
   runtime {
-    docker:"quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0"
-    memory:"5 GB"
+    docker: "quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0"
```
ambrosejcarr: There was some discussion of making dockers parameterized as inputs. Is this something we want to either (1) show in the guide or (2) standardize?

library/tasks/Hisat2.wdl
```diff
@@ -46,35 +46,38 @@ task HISAT2PE {
...
   }
+
   output {
     File logfile = "${output_name}.log"
     File metfile = "${output_name}.hisat2.met.txt"
     File output_bam = "${output_name}.bam"
   }
 }
+
 ## run hisat2 for rsem
```
ambrosejcarr: What's the deal with `#` vs `##` comments?

library/tasks/Picard.wdl
```diff
@@ -95,15 +103,16 @@ task CollectDuplicationMetrics {
        METRICS_FILE="${output_filename}.duplicate_metrics.txt" \
        REMOVE_DUPLICATES=false
   }
+
   runtime {
```
ambrosejcarr: A thought on ordering: Does it matter what order the components of the task are coded in? 

If not, the most coherent order, to me, would be: 
`runtime`, `command`, `output`.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+
+##### New Line Rules #####
+## ** New line before and after `calls`
+## ** New Line between task definitions
+## ** New lines between blocks in tasks
+## ** New line variable declarations that you want to group
+## ** New line between imports and workflow definition
+## ** No new lines at the start or end of any workflow/call/task definition block
+## ** No need for having two new lines anywhere
+## ** No new lines at the start of any blocks
```
ambrosejcarr: I read this as redundant with line 13 above.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+
+##### Spacing Rules #####
+## ** Space before `{` characters
+## ** Space after `,` characters
+## ** Space before and after operators e.g. `+`, `=`
+## ** Space after `:` character in runtime block
+## ** Space after `#` when creating a comment (or `##` if we decide thats our comment definer)
```
ambrosejcarr: We should use # -- it's the minimum definition of a comment, may as well keep it simple and eliminate confusion.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+
+##### Tab vs Space for Indents ######
+## ** Always use spaces? I think whats most frequently used is whenever you want to indent you use 2 spaces but we should decide what works best for us.
```
ambrosejcarr: Always use spaces. 👍

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
ambrosejcarr: I like baking in CPUs and Memory because very often this just means that you're coding in the maximum allotment for the machine type that you require. 

i.e. if you need 30gb of ram and you have a multi-processed tool, you may as well request the 8 cores you get with that. 

Interested to hear counter-examples where this is not productive. I do have some WDLs where things are set variably (e.g. alignment, where ram requirements are a function of genome index size).

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+  File hisat2_ref_name
+  File hisat2_ref_trans_name
+  # samples
+  String stranded
+  String sample_name
+  String output_name
+  File fastq1
+  File fastq2
+
+  call RunHisatQCWDL.RunHisat2Pipeline as QC {  ## ** Should callign a subworkflow be aliased some suffix or prefix so its easier to know when we're dealing with a workflow as opposed to a task?
```
ambrosejcarr: 100% yes. 

If you're calling a workflow, call it as `QC_workflow` -- makes it much, much clearer.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
ambrosejcarr: The below is not very clear from me from the documentation, which has no examples. Can you link me a WDL that uses this? It sounds useful and I'd like for us to utilize it. 

```
Parameter Metadata Section
🐖 Cromwell supported ✅

$parameter_meta = 'parameter_meta' $ws* '{' ($ws* $parameter_meta_kv $ws*)* '}'
$parameter_meta_kv = $identifier $ws* '=' $ws* $string
This purely optional section contains key/value pairs where the keys are names of parameters and the values are string descriptions for those parameters.

Additional requirement: Any key in this section MUST correspond to a parameter in the command line

Metadata Section
🐖 Cromwell supported ✅

$meta = 'meta' $ws* '{' ($ws* $meta_kv $ws*)* '}'
$meta_kv = $identifier $ws* '=' $ws* $string
This purely optional section contains key/value pairs for any additional meta data that should be stored with the task. For example, perhaps author or contact email.
```

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
ambrosejcarr: Is there a _workflow_ meta section? In that case we could also put this information there.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+      hisat2_ref = hisat2_ref_index,
+      hisat2_ref_name = hisat2_ref_name,
+      sample_name = sample_name,
+      output_prefix = output_name + "_qc"
+  }
+
+  ## ** We should also figure out what we want the balance to be between moving stuff to subworkflows or getting rid of
+  ## ** them and moving that functionality up a level.  In this case do we like that QC and DATA are two different workflows
+  ## ** or would we rather move whats in those two wdls into here.  A good question to ask might be "Will we ever run
+  ## ** this subwokflow on its own?" If not, maybe it shouldnt be its own workflow?
```
ambrosejcarr: There are _also_ some stupid corner cases where we need subworkflows, e.g. optimus, due to limitations around nested scatters.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+
+## ** Should comments / descriptions above a task definition be mandatory?
+
+## main pipeline:
+## QC track: HISAT2+Picard
+## this track will produce aligned bam file and a set of QC metrics
+## rsem quantification track: HISAT2+RSEM
+## this track involves hisat2 transcriptome alignment and RSEM gene expression estimation.
+workflow SmartSeq2SingleCell {
+  # load annotation
```
ambrosejcarr: I would put newlines between the comments here to make the blocks more visibly distinct. What do others think?

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+## ** Explanations above the workflow about what inputs it expects, what outputs/metrics it outputs
+## ** What, if any, assumptions are made in the pipeline
+## ** In general any kind of description you want to add above the workflow is nice
+## ** A good example of this can be found in these wdls
+## ** https://github.com/broadinstitute/dsde-pipelines/blob/develop/genomes_in_the_cloud/single_sample/PairedSingleSampleWf.wdl
+## ** https://github.com/broadinstitute/gatk/blob/master/scripts/mutect2_wdl/mutect2.wdl
+
+## ** Should comments / descriptions above a task definition be mandatory?
+
+## main pipeline:
```
ambrosejcarr: Capitalize `Main Pipeline`

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
```diff
@@ -0,0 +1,136 @@
...
+
+##### Tab vs Space for Indents ######
+## ** Always use spaces? I think whats most frequently used is whenever you want to indent you use 2 spaces but we should decide what works best for us.
+
+##### Runtime Attributes ######
+## ** This will change quite a bit i believe when the new spec of WDL comes out (WDL SPEC 3) but we should define all values we care about
+## ** even if they are the default values.  e.g. write `cpu: 1` instead of leaving it out because its the default.
+
+##### Comment Rules ######
+## **  Should comments have one or two `##` before a comment.  Maybe two for workflow or task descriptions and one for any other kind of comment?
```
ambrosejcarr: One. Always. if the comment is defined by WDL as `#`, that's what we should use.

library/tasks/Hisat2.wdl
```diff
@@ -2,19 +2,19 @@
...
   String output_name
   String sample_name
   Int disk_size
+
   command {
     # Note that files MUST be gzipped or the module will not function properly
     # This will be addressed in the future either by a change in how Hisat2 functions or a more
     # robust test for compression type.
-
     set -e
```
ambrosejcarr: Shall we have a section about mandating things like this for bash blocks with multiple commands? I didn't know about this and WDL/Cromwell debugging really depends on having this present, otherwise the error gets bumped to the next task.

library/tasks/Hisat2.wdl
ambrosejcarr: Oh I see, you got to this later.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: there is! and we should def take advantage of it - https://github.com/openwdl/wdl/blob/master/versions/draft-3/SPEC.md#metadata

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: here is an example of  it in a task - https://github.com/openwdl/wdl/blob/master/versions/draft-3/SPEC.md#example-3-runtimemetadata , it would look similar at the workflow level as well

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: good question and honestly I don't have a good answer of what the rules should be other than consistent 😞.  Something to bring up with the team for sure

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: noted as something the team should discuss further since this would be a "mint" thing as opposed to a "wdl" thing.

library/tasks/Hisat2.wdl
jsotobroad: 👍

library/tasks/Picard.wdl
jsotobroad: ordering doesn't matter at all.  I would argue that the `command` block is what people want to see first although if we're now putting in `meta` and `parameter_meta` blocks maybe those push other blocks lower?

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: agreed

library/tasks/Hisat2.wdl
jsotobroad: I purposely stayed away from the command block since that should be covered by bash best practices but if we feel that is important for us we can add it

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: I agree

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
jsotobroad: yup so maybe the two cases we want to allow for `subworkflows` is getting around nested scatters and workflows that would be run on their own

library/tasks/Picard.wdl
ambrosejcarr: I guess I'm thinking about this in terms of dependencies. You put inputs first because it tells you what data and information you have access to. In the same way, I'd argue that runtime comes next because you want to know how much ram/cpu/disk resource you're able to consume before you see the actual command that gets run. 

Then, once you've run something, output comes last.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
ambrosejcarr: I vote for leaving them capitalized, but in cases of jargon acronyms, remove them and spell out the word if possible. (e.g. SS -> SmartSeq, WGS -> WholeGenomeSequencing). 

Conversely, things like BAM, VCF, FASTQ, that all have standards, can be left in.

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
ambrosejcarr: 👍

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
rexwangcc: +1

pipelines/smartseq2_single_sample/SS2SingleSample.wdl
rexwangcc: +100

PR #78, opened 2018-02-06
Ajc remove i7 requirement

ambrosejcarr:
This PR removes the hard requirement that the user pass an index fastq file. The index is not used in any downstream steps in the pipeline and is thought by many to be superfluous. Removing the requirement increases pipeline flexibility.

### DIFF
--- a/library/tasks/Attach10xBarcodes.wdl
+++ b/library/tasks/Attach10xBarcodes.wdl
@@ -3,26 +3,26 @@ task Attach10xBarcodes {
   # attaches barcodes found in r1 and i1
 
   File r1  # forward fastq file; contains umi, cell barcode
-  File i1  # index fastq file; contains sample barcode
   File u2  # reverse bam file; contains alignable genomic information
+  File? i1  # optional, index fastq file; contains sample barcode
   File whitelist  # 10x genomics cell barcode whitelist for 10x V2
 
   # estimate disk requirements
   Float bam_size = size(u2, "G")
-  Float fastq_size = size(r1, "G") + size(i1, "G")
+  Float fastq_size = size(r1, "G")  + if (defined(i1)) then size(i1, "G") else 0
   Int estimated_required_disk = ceil(fastq_size + bam_size * 2.5)
 
   command {
     Attach10xBarcodes \
       --r1 "${r1}" \
-      --i1 "${i1}" \
+      ${"--i1 " + i1} \
       --u2 "${u2}" \
       --output-bamfile barcoded.bam \
       --whitelist "${whitelist}"
   }
   
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-python3-scientific:0.1.5"
+    docker: "quay.io/humancellatlas/secondary-analysis-sctools:0.1.6"
     cpu: 2
     memory: "7.5 GB"
     disks: "local-disk ${estimated_required_disk} HDD"
--- a/pipelines/optimus/Optimus.wdl
+++ b/pipelines/optimus/Optimus.wdl
@@ -12,34 +12,60 @@ import "AlignTagCorrectUmis.wdl" as AlignTagCorrectUmis
 # format.
 workflow Optimus {
 
-  Array[Array[File]] fastq_inputs  # array of arrays of matched fastqs [ [r1, r2, i1], ... ]
-  File whitelist  # 10x genomics cell barcode whitelist for 10x V2
+  # Sequencing data inputs (fastq)
+  Array[File] r1 # forward read, contains cell barcodes and molecule barcodes
+  Array[File] r2 # reverse read, contains cDNA fragment generated from captured mRNA
+  Array[File]? i1 # (optional) index read, for demultiplexing of multiple samples on one flow cell.
+
+  String sample_id  # name of sample matching this file, inserted into read group header
+
+  # organism reference parameters
   File tar_star_reference  # star reference
   File annotations_gtf  # gtf containing annotations for gene tagging
   File ref_genome_fasta  # genome fasta file
-  String sample_id  # name of sample matching this file, inserted into read group header
-  String fastq_suffix = ""  # when running in green box, need to add ".gz" for picard to detect
 
-  # this scatters matched [r1, r2, i1] fastq arrays
-  scatter (fastqs in fastq_inputs) {
+  # 10x v2 parameters
+  File whitelist  # 10x genomics cell barcode whitelist for 10x V2
+
+  # environment-specific parameters
+  String fastq_suffix = ""  # when running in green box, need to add ".gz" for picard to detect
+  Array[Int] indices = range(length(r1)) # this scatters matched [r1, r2, i1] fastq arrays
+
+  scatter (index in indices) {
     call fq2bam.FastqToUBam {
       input:
-        fastq_file = fastqs[1],
+        fastq_file = r2[index],
         sample_id = sample_id,
         fastq_suffix = fastq_suffix
     }
 
-    call attach.Attach10xBarcodes {
-      input:
-        r1 = fastqs[0],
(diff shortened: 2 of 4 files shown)


ambrosejcarr:
@jsotobroad 

I'm running into the following error on this PR: 

`Unable to build WOM node for WdlWorkflowCall 'target': Unable to build WOM node for Scatter '$scatter_0': Unable to build WOM node for If '$if_0': Unable to build WOM node for WdlTaskCall 'attach_barcodes': Invalid indexing target. You cannot index a value of type 'Array[File]?'`

... can I not index an optional Array[File]? That seems like it more or less invalidates the usefulness of the object, so I assume I'm doing something wrong. Can you take a look?

jsotobroad:
yup unfortunately you have to do `Array[File] make_non_optional = select_first([i1])` to get a non optional array, you can then index `make_non_optional`

ambrosejcarr:
@jsotobroad could you do a real review of this? your suggestion fixed the problem. :)

library/tasks/Attach10xBarcodes.wdl
```diff
@@ -3,26 +3,26 @@ task Attach10xBarcodes {
...
-      --i1 "${i1}" \
+      ${"--i1 " + i1} \
       --u2 "${u2}" \
       --output-bamfile barcoded.bam \
       --whitelist "${whitelist}"
   }
   
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-python3-scientific:0.1.5"
+    docker: "quay.io/humancellatlas/secondary-analysis-sctools:ajc-remove-i7-requirement-517"
```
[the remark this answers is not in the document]
jsotobroad: you said at standup that it moved from secondary analysis to skylab i believe or vice versa?  We shoudl just make sure the new name makes sense

pipelines/optimus/Optimus.wdl
```diff
@@ -12,34 +12,60 @@ import "AlignTagCorrectUmis.wdl" as AlignTagCorrectUmis
...
+        input:
+          r1 = r1[index],
+          i1 = non_optional_i1[index],
+          u2 = FastqToUBam.bam_output,
+          whitelist = whitelist
+      }
     }
 
+    # if the index is not passed, proceed without it.
+    if (!defined(i1)) {
```
jsotobroad: I couldn't think of a better way to do this 😢

library/tasks/Attach10xBarcodes.wdl
```diff
@@ -3,26 +3,26 @@ task Attach10xBarcodes {
   # attaches barcodes found in r1 and i1
```
jsotobroad: I see that `\n` at the top!

library/tasks/Attach10xBarcodes.wdl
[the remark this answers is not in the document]
ambrosejcarr: This is the name of the branch, which is used during testing. Given that tests have passed, I think the first step is to finalize the sctools PR (also linked in the ticket) and then push a version tag to that repo which will auto-build it. 

This should be incorporated before we merge.

pipelines/optimus/Optimus.wdl
ambrosejcarr: 👍

test/optimus/pr/test_inputs.json
```diff
@@ -3,18 +3,19 @@
...
-    ]
+  "TestOptimusPR.r1": [
+    "gs://broad-dsde-mint-dev-teststorage/10x/demo/fastqs/pbmc8k_S1_L007_R1_001.fastq.gz",
+    "gs://broad-dsde-mint-dev-teststorage/10x/demo/fastqs/pbmc8k_S1_L007_R1_001.fastq.gz"
   ],
+  "TestOptimusPR.r2": [
+    "gs://broad-dsde-mint-dev-teststorage/10x/demo/fastqs/pbmc8k_S1_L007_R2_001.fastq.gz",
+    "gs://broad-dsde-mint-dev-teststorage/10x/demo/fastqs/pbmc8k_S1_L007_R2_001.fastq.gz"
+  ],
+  "TestOptimusPR.i1": [
```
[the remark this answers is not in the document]
ambrosejcarr: I'm going to push a commit wherein we remove those i1 file inputs and see where the PR test fails. I expect that the bam file should no longer look the same, but that the pipeline should complete execution and only the _checker_ should fail. 

If this occurs, I think we're good to merge.

library/tasks/Attach10xBarcodes.wdl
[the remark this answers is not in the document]
jsotobroad: Gotcha thanks for the explanation

test/optimus/pr/test_optimus_PR.wdl
```diff
@@ -13,7 +13,10 @@ workflow TestOptimusPR {
   String expected_picard_metrics_hash
 
   # Optimus inputs
-  Array[Array[File]] fastq_inputs  # array of arrays of matched fastqs [ [r1, r2, i1], ... ]
+  Array[File] r1
+  Array[File] r2
+  Array[File] i1
```
jsotobroad: I think you'll have to make this optional if you want to test not passing in an `i1` file.  Does our testing infrastructure allow for testing a workflow against multiple input jsons / is this something we even want to do?

ambrosejcarr: It should allow this, yes. I'm not sure we'd want to put in this use case.

ambrosejcarr:
Good call, I will bring it up at stand-up tomorrow. 👍 

I doubt I'll merge this before then.

library/tasks/Attach10xBarcodes.wdl
jishuxu: I remember there was some conversation on docker tagging and docker auto-build from each branch. I don't think we settle on how to do this yet.

pipelines/optimus/Optimus.wdl
jishuxu: i guess if we 100% sure we won't need I1,we can just simple take I1 out of the pipeline.

pipelines/optimus/Optimus.wdl
```diff
@@ -12,34 +12,60 @@ import "AlignTagCorrectUmis.wdl" as AlignTagCorrectUmis
 # format.
 workflow Optimus {
 
-  Array[Array[File]] fastq_inputs  # array of arrays of matched fastqs [ [r1, r2, i1], ... ]
-  File whitelist  # 10x genomics cell barcode whitelist for 10x V2
+  # Sequencing data inputs (fastq)
+  Array[File] r1 # forward read, contains cell barcodes and molecule barcodes
```
jishuxu: Array[File] is much better!

ambrosejcarr:
Sorry I guess I wasn't clear about that Jishu -- there are lots of cases wherein one might pass those files, they are just not _currently_ consumed by the pipeline in any way.

jishuxu:
we might need to do some test with various input, such as w/o I1 reads. Just want to make sure pipeline run through, nothing fancy. As soon as these tests run through, it is fine with me to merge.

ambrosejcarr:
jenkins retest

ambrosejcarr:
@jsotobroad @jishuxu I've confirmed that making the i1 file optional and removing the i1 inputs runs through the Optimus workflow. See: 

https://cromwell.mint-dev.broadinstitute.org/api/workflows/v1/239f572c-7b8f-4920-8bd8-236f6dbbbd9b/metadata?expandSubWorkflows=false

Note that this is the PR workflow, so the overall workflow fails the checker task. However, "target" succeeds and the output looks good -- it's just missing i1 tags and therefore fails the md5. Given this, I think that reverting the input json to pass the i1 file will be good to merge.

PR #79, opened 2018-02-08
added set -e, boosted memory requirements for mixed genomes

ambrosejcarr:
Minor PR to `set -e` and boost memory requested for mixed genomes. We don't use this very often, so I think the increased memory is a good idea to make sure it always works when we call it. 

I was getting failures at 60 GB.

### DIFF
--- a/library/tasks/StarMkref.wdl
+++ b/library/tasks/StarMkref.wdl
@@ -7,7 +7,10 @@ task StarMkref {
   String star_docker_image = "quay.io/humancellatlas/secondary-analysis-star:v0.2.2-2.5.3a-40ead6e"
   
   command {
+    set -e
+
     mkdir genome
+
     STAR \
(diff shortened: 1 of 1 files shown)


library/tasks/StarMkref.wdl
```diff
@@ -7,7 +7,10 @@ task StarMkref {
   String star_docker_image = "quay.io/humancellatlas/secondary-analysis-star:v0.2.2-2.5.3a-40ead6e"
```
jishuxu: if you want to put docker name as variable, you'd better pass them in the input file, right?  we can put all docker names in a json files, such as docker.json and pass it to wdl, is it possible?

ambrosejcarr: Heh, I forgot I'd done this. Yea... we could change this across our wdls.

PR #81, opened 2018-02-12
Cw add pullapprove 554

rexwangcc:
Previously we had soft team agreement that each Skylab PR requires approvals from both a comp-bio and a sde.  This PR adds pullapprove plugin to Skylab to make the team agreement solid. This may slow down the development efficiency a little bit at the beginning but will provide better safety and code quality in the future. 

~Also, note that `skip-review` label can be added to an emergent PR to skip the pull approval.~

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,0 +1,40 @@
+# You *must* specify "version: 2"
+version: 2
+
+requirements:
+  signed_off_by:
+    required: false
+
+always_pending:  # A pull request matching any of these conditions will always have a "pending" status for this group.
+  title_regex: '(WIP|wip)'
+  explanation: 'Work in progress...'
+
+# Group settings to apply to all groups by default, optionally being overridden later
+group_defaults:
+  reset_on_push: # Whether or not approval statuses get reset to "pending" when new commits are pushed to a PR
+    enabled: false # Set to false for efficiency
+  reset_on_reopened: # Whether or not approval statuses get reset to "pending" when a pull request is re-opened
+    enabled: true
+  author_approval: # Whether or not the author must approve the pull request.
+    required: false
+    ignored: true
+  approve_by_comment:
+    enabled: false
+  reject_value: -1 # Essentially allows rejections to cancel out approvals
+
+# Groups of reviewers and their respective settings
+groups:
+  computational-biologists:
+    required: 1 # require approval from at least one computational biologist
+    users:
+      - ambrosejcarr
+      - jishuxu
+      - mckinsel
+
+  software-engineers:
+    required: 1 # require approval from at least one software engineer
+    users:
+      - dshiga
+      - jsotobroad
+      - samanehsan
+      - rexwangcc


.pullapprove.yml
```diff
@@ -0,0 +1,56 @@
...
+  reset_on_push: # Whether or not approval statuses get reset to "pending" when new commits are pushed to a PR
+    enabled: false # Set to false for efficiency
+  reset_on_reopened: # Whether or not approval statuses get reset to "pending" when a pull request is re-opened
+    enabled: true
+  author_approval: # Whether or not the author must approve the pull request.
+    required: false
+    ignored: true
+  conditions:
+    branches:
+      - master  # Only require reviews when PR is merging into master
```
dshiga: This situation hasn't really come up yet, but I'd rather err on the side of being too strict for now. We can always open it up later if needed.

.pullapprove.yml
```diff
@@ -0,0 +1,56 @@
...
+    labels:
+      exclude:
+        - skip-review  # don't need to review if it has the "skip-review" label, designed for "fire-fighting" PRs
+    files:
+      include:
+        - '*.md'
+      exclude: # don't need to review if the PR only changes these files
+        - 'README.md'
```
dshiga: I think readme changes are important to review as well, so I'd suggest not excluding them here.

.pullapprove.yml
```diff
@@ -0,0 +1,56 @@
...
+  approve_by_comment:
+    enabled: true
+    approve_regex: '(^[Aa]pproved)|(^LGTM)'  # overrides the default '^:\+1:'
```
dshiga: I prefer not to do this. I'd rather we have exactly one way that a PR gets approved, to reduce any possible ambiguity. If that turns out to be too onerous we can always revisit.

.pullapprove.yml
```diff
@@ -0,0 +1,56 @@
...
+
+# Groups of reviewers and their respective settings
+groups:
+  computational-biologists:
+    required: 1 # require approval from at least one computational biologist
+    users:
+      - ambrosejcarr
+      - jishuxu
+      - mckinsel
+      - timothytickle
```
dshiga: Although @kbergin and @TimothyTickle have lots of expertise to offer, given their roles on the team, they have not typically been giving official approvals on PRs and I don't expect them to in future. I think we should remove them from the lists of required reviewers. This wouldn't stop them from reviewing PRs but means we would still require at least one engineer and one comp bio from the team to review, in addition to them. Again, we could revisit in future if it turns out there are PRs where it makes sense for them to be the official approvers.

.pullapprove.yml
```diff
@@ -0,0 +1,56 @@
...
+    enabled: true
+  author_approval: # Whether or not the author must approve the pull request.
+    required: false
+    ignored: true
+  conditions:
+    branches:
+      - master  # Only require reviews when PR is merging into master
+    labels:
+      exclude:
+        - skip-review  # don't need to review if it has the "skip-review" label, designed for "fire-fighting" PRs
```
dshiga: This is another thing I feel we should remove for now and add back in later if we find it to be important.

rexwangcc:
Thanks for the review! Will rebase on master branch to fix the Circle CI.

rexwangcc:
Thanks for all the reviews, will merge this PR in now.

PR #83, opened 2018-02-13
Ajc adjust star to support mixed genomes 517

ambrosejcarr:
Minor adjustments to the TagGeneExon task to make it run on larger genomes (memory usage scales with genome size)

### DIFF
--- a/library/tasks/StarAlignBamSingleEnd.wdl
+++ b/library/tasks/StarAlignBamSingleEnd.wdl
@@ -8,6 +8,7 @@ task StarAlignBamSingleEnd {
   Int input_size = ceil(size(bam_input, "G"))
   Int reference_size = ceil(size(tar_star_reference, "G"))
   Int estimated_disk_required =  ceil(input_size * 2.2 + reference_size * 2)
+  Int estimated_memory_required = reference_size + 6
 
   command {
     # prepare reference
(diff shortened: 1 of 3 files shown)


library/tasks/StarAlignBamSingleEnd.wdl
```diff
@@ -8,6 +8,7 @@ task StarAlignBamSingleEnd {
   Int input_size = ceil(size(bam_input, "G"))
   Int reference_size = ceil(size(tar_star_reference, "G"))
   Int estimated_disk_required =  ceil(input_size * 2.2 + reference_size * 2)
+  Int estimated_memory_required = reference_size + 6
```
rexwangcc: Which is larger? 30 GB or `reference_size + 6` GB?

rexwangcc:
You may want to update this branch so that you could pass the Circle CI test.

library/tasks/StarAlignBamSingleEnd.wdl
ambrosejcarr: Depends on reference_size. :)

mckinsel:
Hey I've got a somewhat meta-question about tuning these disk size parameters. It seems like we're being pretty conservative when setting disk size, trying to cut it relatively close. But I had thought that cost of disk is pretty low, and the cost of having an engineer investigate a job that failed because it ran out of disk is very high. Does it make sense for us to ease up a little on the disk sizes? It might matter a lot once we start seeing production volumes.

ambrosejcarr:
So I think you're right, but incidentally we managed to successfully run an absolutely huge dataset of 25x hiseqs in a single bundle, so maybe we already accomplished this (?)

... this particular example is a weird case where we had a _memory_ scaling issue that wasn't expected because memory usage scales with genome size.

PR #84, opened 2018-02-15
Use public inputs in example files

dshiga:
Previously, the example inputs files were pointing to a private bucket in our dev project. This PR changes them to point to inputs in a public bucket instead.

See:
https://elastc.com/c/QN-0A7Mr/539-remove-environment-specific-params-from-example-inputs

Related PR: https://github.com/HumanCellAtlas/pipeline-tools/pull/18

PR #86, opened 2018-02-20
Switch Optimus inputs to public bucket

mckinsel:


test/optimus/pr/test_inputs.json
```diff
@@ -4,20 +4,20 @@
   "TestOptimusPR.expected_matrix_summary_hash": "dd513351d4e7688c97f7bf902ba2876f",
   "TestOptimusPR.expected_picard_metrics_hash": "7b7be5c9a2236920ca09f05811dca6d5",
   "TestOptimusPR.r1": [
-    "gs://broad-dsde-mint-dev-teststorage/10x/demo/fastqs/pbmc8k_S1_L007_R1_001.fastq.gz",
-    "gs://broad-dsde-mint-dev-teststorage/10x/demo/fastqs/pbmc8k_S1_L007_R1_001.fastq.gz"
+    "gs://hca-dcp-mint-test-data/optimus/fastqs/pbmc8k_S1_L007_R1_001.fastq.gz",
```
samanehsan: I think the test inputs should be the same as the inputs here for consistency: https://github.com/HumanCellAtlas/skylab/pull/84/files#diff-7787cfa08c74e3392fe187d33f67dca1R4, although I'm not sure whether we want to use the data in the Optimus or 10x folder (I didn't realize there were both)!

mckinsel:
Hey thanks for the comments. So there are files in two places right now, `10x/demo` and `optimus`. It seems like they should only be in one place, but which one?

ambrosejcarr:
My opinion is that the files that are the current targets of the PR tests should reside in `10x/demo` and be used as inputs for all of our 10x v2 testing.

TimothyTickle:
Often in my experience, "demo" files in software repos are toy data sets that are super fast and help people get up and running, minimal data sets that demonstrate install and functionality pass. I am not sure that is the kind of file you want a PR based on. Last time we spoke it seemed the team was advocating inputs that were closer to reality. These may be two different data.

ambrosejcarr:
I understood that we were talking about a "PR test" which would run rapidly to test that results are identical, and a "scientific test" which would run frequently but not on PR, which would provide a baseline against which we could compare the outputs. 

These data files are appropriate for the PR test, as defined above, but not the scientific test.

mckinsel:
jenkins retest

test/optimus/pr/test_inputs.json
```diff
@@ -4,20 +4,20 @@
...
+    "gs://hca-dcp-mint-test-data/10x/demo/fastqs/pbmc8k_S1_L007_I1_001.fastq.gz"
   ],
-  "TestOptimusPR.whitelist": "gs://broad-dsde-mint-dev-teststorage/10x/whitelist/737K-august-2016.txt",
-  "TestOptimusPR.tar_star_reference": "gs://broad-dsde-mint-dev-teststorage/demo/star.tar",
+  "TestOptimusPR.whitelist": "gs://hca-dcp-mint-test-data/10x/whitelist/737K-august-2016.txt",
+  "TestOptimusPR.tar_star_reference": "gs://hca-dcp-mint-test-data/optimus/star.tar",
   "TestOptimusPR.sample_id": "pbmc8k_test",
-  "TestOptimusPR.annotations_gtf": "gs://broad-dsde-mint-dev-teststorage/reference/hg19_ds/GSM1629193_hg19_ERCC.gtf.gz",
-  "TestOptimusPR.ref_genome_fasta": "gs://broad-dsde-mint-dev-teststorage/demo/chr21.fa"
+  "TestOptimusPR.annotations_gtf": "gs://hca-dcp-mint-test-data/optimus/GSM1629193_hg19_ERCC.gtf.gz",
```
samanehsan: @mckinsel is this the same file as `gs://hca-dcp-mint-test-data/reference/demo/hg19_ds/GSM1629193_hg19_ERCC.gtf.gz`? That's what is listed in the example inputs for the adapter pipeline, and I'm not sure if they're supposed to be the same: https://github.com/HumanCellAtlas/skylab/blob/master/pipelines/optimus/example_test_inputs.json#L17

test/optimus/pr/test_inputs.json
```diff
@@ -4,20 +4,20 @@
...
+  "TestOptimusPR.ref_genome_fasta": "gs://hca-dcp-mint-test-data/optimus/chr21.fa"
```
samanehsan: Same as above, the adapter input file is using "gs://hca-dcp-mint-test-data/reference/demo/chr21.fa". Just wondering if that's a duplicate file, or if they are supposed to be different here.

test/optimus/pr/test_inputs.json
mckinsel: Pretty sure it's the same, I'll change the path.

test/optimus/pr/test_inputs.json
mckinsel: This also should be the same, as well as the star.tar. I'll change both paths.

PR #87, opened 2018-02-25
Adjust parallelism in Optimus

jsotobroad:
We want to make sure all reads for a given cell barcode are mark duplicated together.


ambrosejcarr:
The code changes here all look good to me. I have some follow-up comments on the structure: 

1. I'm not sure if we need to lead with a sort or not. SplitBamByCellBarcode will not disrupt the sort order, but it could make sense to do the sort afterwards to speed things up. 
  1. Conversely, If we decide that we want to do the sort up-front, we should make a ticket to modify SplitSamByCellBarcode to have an `assume-sorted` tag -- right now it does not do that, and it could be _much_ faster if it assumed sorting (or pre-sorted, heh...). 
2. We should completely remove the sub-workflow now that we don't have a double-scatter.

jsotobroad:
hash updated, thanks

library/tasks/MergeSortBam.wdl
```diff
@@ -1,18 +1,17 @@
...
-    input_line=$(python -c "print(' INPUT='.join('${sep=' ' bam_inputs}'.replace('[', '').replace(']', '').replace(',', '').split()))")
-
-    echo $input_line
     java -Dsamjdk.compression_level=${compression_level} -Xms7000m -jar /usr/gitc/picard.jar \
       MergeSamFiles \
       USE_THREADING=true \
-      SORT_ORDER=coordinate \
-      INPUT=$input_line \
+      SORT_ORDER=${sort_order} \
+      INPUT=${sep=' INPUT=' bam_inputs} \
```
samanehsan: Won't this insert ` INPUT=` in between each item in the `bam_inputs` array instead of joining the items in the array with ` INPUT=` as a prefix like it was previously?

samanehsan:
Please remove all references to `AlignTagCorrectUmis` now that the file is deleted. I believe these are the last two places but definitely confirm! 
- https://github.com/HumanCellAtlas/skylab/blob/357e165113d8929c35c84842ccb3ef0da09fa68f/pipelines/optimus/example_test_outputs.json

- https://github.com/HumanCellAtlas/skylab/blob/ce5fb30dd214b54a50956c6ed2cf5707f43cb201/test/test_pr_tests.ipynb

samanehsan:
This looks great @jsotobroad! I just have a couple of comments for you!

library/tasks/MergeSortBam.wdl
jsotobroad: I believe the python was doing the same thing as this `sep` command.  Splitting the `Array[Array[File]] bam_inputs` into a string with spaces between each element by `.replace()`ing all `[`, `]`, `,` characters with empty characters.  It then `.split()`s that string by spaces to get a python array and then `.join()`s that array with " INPUT=".  Because I was able to change the `Array[Array[File]] bam_inputs` input to `Array[File] bam_inputs` it was simpler to use wdls `sep` to accomplish the same thing.

jsotobroad:
Thanks for those catches @samanehsan, hopefully we got em all!

library/tasks/MergeSortBam.wdl
samanehsan: Oh I see now! I misunderstood what it was doing before. Thanks for clarifying!

samanehsan:
Also please rebase your commits into just one :)

rexwangcc:
jenkins retest

PR #89, opened 2018-03-07
[Do Not Merge] Try applying style guide

jsotobroad:
This PR is to get feedback on how we want to allow users to change runtime arguments through task inputs.  There are 3 ways I have tried that I will point out in the PR.  If there are other suggestions please let us know as well.

I also checked to see if `meta` or `parameter_meta` blocks could be the first thing in a task and it turns out you can.  The red squiggly line we saw during the style meeting must have been a syntax parser bug or something.  With that being available to us, the same three tasks with the different runtime versions also have three different orders of the `inputs`, `meta`, and `parameter_meta` blocks in case anyone wanted to change their mind as to what we had decided, which was.

`input`
`meta`
`parameter_meta`


library/tasks/HISAT2.wdl
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  File hisat2_ref
+  File fq1
+  File fq2
+  String ref_name
+  String output_name
+  String sample_name
+
+  # runtime optional arguments
```
jsotobroad: The idea for this process is that there are explicit optional arguments for the task that then get converted to non optional values though some `select_first([optional_value, default_value])` expression.  Things I like about this is that its very explicit as to what is expected for the user to put an input to if they wish to change a runtime value.  I don't like how much boilerplate code this adds to every task though.

library/tasks/HISAT2.wdl
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  File hisat2_ref
+  File fq1
+  File fq2
+  String ref_name
+  String output_name
+  String sample_name
+
+  # runtime values
```
jsotobroad: this version only has non optional values with the default values set in by default.  I like this because its way less boilerplate and it looks cleaner.  I think it could get a little confusing as to whether or not users will know they can/should overwrite these or not but I think if we put the right descriptions in the `parameter_meta` we might be able to make it more clear.

library/tasks/HISAT2.wdl
```diff
@@ -0,0 +1,301 @@
...
+}
+
+task HISAT2SingleEnd {
+  File hisat2_ref
+  File fq
+  String ref_name
+  String output_name
+  String sample_name
+
+  # runtime values
```
jsotobroad: This version is the same as the previous one except all the values are optional.  This explicitly tells the user that they can optionally set these values but they do have a default.  Downside to this is that I don't really think optionals are meant to be used this way and you have to add a `select_first()` where ever you do plan to use these values.

library/tasks/HISAT2.wdl
meganshand: It seems that the only additional code in this task is this block:
```wdl
 String? opt_docker
 Int? opt_memory_gb
 Int? opt_cpu
 Int? opt_disk
 Int? opt_preemptible
```
Is that right? If that's the only additional boilerplate over the following version, then this option seems clearest to the user in my opinion. That said, I haven't used `parameter_meta` before, so maybe I'm biased.

library/tasks/HISAT2.wdl
meganshand: For this version, should the `parameter_meta` contain the optional variables such as `opt_docker` instead of `docker`?

library/tasks/HISAT2.wdl
jsotobroad: yup those 5 inputs are the "extra boilerplate". I wasn't sure if 5 lines per task were too much but good to hear that you think its acceptable!

library/tasks/HISAT2.wdl
jsotobroad: it should not, i didnt do a good job at fleshing out the different version fully, will correct this 👍

PR #91, opened 2018-03-15
Update .pullapprove.yml

rexwangcc:
Add meganshand to the comp-bio group.


dshiga:
Oops, I shouldn't have approved - there's some problem with the pullapprove yaml apparently?

rexwangcc:
The int test failed at submission step, which apparently is not related to this PR. Will merge the PR in now.

PR #99, opened 2018-04-01
Improve SmartSeq2 portability

mckinsel:
This PR has a few of changes to the SS2 test and pipeline that help it run better in other environments.

### DIFF
--- a/library/tasks/HISAT2.wdl
+++ b/library/tasks/HISAT2.wdl
@@ -8,7 +8,7 @@ task HISAT2PairedEnd {
 
   # runtime values
   String docker = "quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0"
-  Int machine_mem_mb = 5000
+  Int machine_mem_mb = 15000
   Int cpu = 4
   # use provided disk number or dynamically size on our own, 10 is our zipped fastq -> bam conversion with 50GB of additional disk
   Int disk = ceil((size(fastq1, "GB") + size(fastq2, "GB") * 10) + size(hisat2_ref, "GB") + 50)
@@ -53,7 +53,7 @@ task HISAT2PairedEnd {
         FQ2=${fastq2}
     fi
 
-    tar -zxvf "${hisat2_ref}"
+    tar --no-same-owner -xvf "${hisat2_ref}"
 
     # run HISAT2 to genome reference with dedault parameters
     # --seed to fix pseudo-random number and in order to produce deterministics results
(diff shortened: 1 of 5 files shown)


library/tasks/RSEM.wdl
```diff
@@ -29,7 +29,7 @@ task RSEMExpression {
   command {
     set -e
   
-    tar -xvf ${rsem_genome}
+    tar --no-same-owner -xvf ${rsem_genome}
```
jishuxu: `rsem` and `hisat2` indexing files are `tar.gz`. is it okay to just use `tar -xvf` to extract `.tar.gz` file?

rexwangcc: Good point here! If they are gzip files, `-z` will be necessary here I guess

mckinsel: You actually don't need the `-z` for gnu tar. See here: https://www.gnu.org/software/tar/manual/tar.html#SEC135

"Reading compressed archive is even simpler: you don't need to specify any additional options as GNU tar recognizes its format automatically."

Also, I think the PR tests would have failed.

rexwangcc: I see, thanks!

jishuxu: good to know this, thanks!

library/tasks/HISAT2.wdl
```diff
@@ -8,7 +8,7 @@ task HISAT2PairedEnd {
 
   # runtime values
   String docker = "quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0"
-  Int machine_mem_mb = 5000
+  Int machine_mem_mb = 15000
```
jishuxu: why do you request `15G` ?

mckinsel: These tasks request 4 cores and 5GB of memory. On GCP, this should get you an n1-standard-4, which actually has 15GB of memory. On AWS, you might get a c4.xlarge, which has 7.5GB of memory.

Some of the tasks appear to actually require between 7.5 and 15GB of memory, so they succeed on GCP but fail on AWS even though both platforms provide an instance that meets the declared requirements.

jishuxu: Yes, it is fair. But 15G is actually overkill for this task. On average, HISAT2 uses ~8G so I would like to request 10G instead 15G here. Does `10G` work with AWS?

mckinsel: It does, but just like GCP you'll get an instance with 15 or 16GB of memory. I'm a little reluctant to set it lower than what we're actually using in production. If 8GB is average but say 10% of jobs use >10GB, then we'd never notice problems in the Green Box, but other users would see lots of failures.

PR #110, opened 2018-05-22
scientific bam tests

ambrosejcarr:
- Adds a scientific test comparing alignments in the optimus aligned/tagged bam output with expectations built from Cell Ranger.


rexwangcc:
retest this please

test/optimus/scientific/ScientificTests.wdl
```diff
@@ -1,26 +1,70 @@
...
+  # Compare the tagged/aligned bam generated by Optimus to expected values derived from running Cell Ranger on the same data"
 
-# note that this is not a great test because it localizes the .bam file
-# however it is a good placeholder while we put together the scientific tests
-# themselves, as these files will be relatively small in the testing data.
-task TestBamRecordNumber {
   File bam
-  Int expected_records
+
+  Int expected_records = 3055752
```
samanehsan: Could you pass in these expected values from the cell ranger wdl running on the same data instead of hard-coding them here?

test/optimus/scientific/ScientificTests.wdl
```diff
@@ -1,26 +1,70 @@
+task TestBam {
+
+  # Compare the tagged/aligned bam generated by Optimus to expected values derived from running Cell Ranger on the same data"
```
samanehsan: I think wdl has a `description` block you can put this in

test/optimus/scientific/ScientificTests.wdl
```diff
@@ -1,26 +1,70 @@
...
+
+    python3 <<CODE
+
+    import pysam
+    import sys
+
+    bamfile = '${bam}'
+    bam = pysam.AlignmentFile(bamfile)
+
+    chr1_records = sum(1 for _ in bam.fetch(contig='1'))
```
samanehsan: Could you simplify this by getting the length of bam.fetch(contig='1')? Same for `chr2_records`

test/optimus/scientific/ScientificTests.wdl
```diff
@@ -1,26 +1,70 @@
...
+    chr2_expected = ${chr2_expected}
+    if chr2_records != chr2_expected:
+        sys.stderr.write('chr2 observed records %d != expected records %d' % (chr2_records, chr2_expected))
+        sys.exit(1)
+
+    unmapped_expected = ${unmapped_expected}
+    if unmapped_records != unmapped_expected:
+        sys.stderr.write('unmapped observed records %d != expected records %d' % (unmapped_records, unmapped_expected))
+        sys.exit(1)
+
```
samanehsan: What do you think about exiting at the end if there are any errors, instead of after each one? If there are multiple errors you could then see all of them instead of failing after the first error.

test/optimus/scientific/ScientificTests.wdl
ambrosejcarr: Ah, which is indecipherably different from meta, which I already tried? 

**sigh**.

test/optimus/scientific/ScientificTests.wdl
ambrosejcarr: Generate has no length, was the error I got. Good thought, though!

test/optimus/scientific/ScientificTests.wdl
ambrosejcarr: Yep, good idea.

test/optimus/scientific/ScientificTests.wdl
ambrosejcarr: I am not allowed to check in the cell ranger wdl yet, but eventually the plan is to calculate these on the fly, yep. 👍

PR #114, opened 2018-06-07
Add automated retries to ss2 tasks (790)

dshiga:
See https://elastc.com/c/dkauyyIF/790-make-ss2-have-cromwell-retries
Cromwell [documentation on maxRetries](http://cromwell.readthedocs.io/en/develop/RuntimeAttributes/#maxretries)
and related PRs
https://github.com/HumanCellAtlas/pipeline-tools/pull/51
https://github.com/HumanCellAtlas/lira/pull/82

- [ ] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/library/tasks/HISAT2.wdl
+++ b/library/tasks/HISAT2.wdl
@@ -13,6 +13,7 @@ task HISAT2PairedEnd {
   # use provided disk number or dynamically size on our own, 10 is our zipped fastq -> bam conversion with 50GB of additional disk
   Int disk = ceil((size(fastq1, "GB") + size(fastq2, "GB") * 10) + size(hisat2_ref, "GB") + 50)
   Int preemptible = 5
+  Int max_retries = 0
 
   meta {
     description: "HISAT2 alignment task will align paired-end fastq reads to reference genome."
@@ -30,6 +31,7 @@ task HISAT2PairedEnd {
     cpu: "(optional) the number of cpus to provision for this task"
     disk: "(optional) the amount of disk space (GB) to provision for this task"
     preemptible: "(optional) if non-zero, request a pre-emptible instance and allow for this number of preemptions before running the task on a non preemptible machine"
+    max_retries: "(optional) retry this number of times if task fails -- use with caution, see skylab README for details"
   }
 
   command {
@@ -79,6 +81,7 @@ task HISAT2PairedEnd {
     disks: "local-disk ${disk} HDD"
     cpu: cpu
     preemptible: preemptible
+    maxRetries: max_retries
   }
 
   output {
@@ -103,6 +106,7 @@ task HISAT2RSEM {
   # use provided disk number or dynamically size on our own, 10 is our zipped fastq -> bam conversion with 50GB of additional disk
   Int disk = ceil((size(fastq1, "GB") + size(fastq2, "GB") * 10) + size(hisat2_ref, "GB") + 50)
   Int preemptible = 5
+  Int max_retries = 0
 
   meta {
     description: "This HISAT2 alignment task will align paired-end fastq reads to transcriptome only. "
@@ -120,6 +124,7 @@ task HISAT2RSEM {
     cpu: "(optional) the number of cpus to provision for this task"
     disk: "(optional) the amount of disk space (GB) to provision for this task"
     preemptible: "(optional) if non-zero, request a pre-emptible instance and allow for this number of preemptions before running the task on a non preemptible machine"
+    max_retries: "(optional) retry this number of times if task fails -- use with caution, see skylab README for details"
   }
 
   command {
@@ -178,6 +183,7 @@ task HISAT2RSEM {
     disks: "local-disk ${disk} HDD"
     cpu: cpu
     preemptible: preemptible
+    maxRetries: max_retries
   }
 
   output {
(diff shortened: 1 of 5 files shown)


library/tasks/HISAT2.wdl
```diff
@@ -13,6 +13,7 @@ task HISAT2PairedEnd {
   # use provided disk number or dynamically size on our own, 10 is our zipped fastq -> bam conversion with 50GB of additional disk
   Int disk = ceil((size(fastq1, "GB") + size(fastq2, "GB") * 10) + size(hisat2_ref, "GB") + 50)
   Int preemptible = 5
+  Int max_retries = 0
```
rexwangcc: I'm not sure if understand why we set 0 here. I guess the upper level(workflow level) matx_retries will set it to 1 if it's available, but it will be unclear to me whether this value is 0 or 1 if we set values at both places. So how about using something like (I'm not sure if the following syntax is correct):
```
Int max_retries = select_first([max_retries, 0])
```
So it guarantees to attempt to use the workflow value first.

dshiga: This is already the way Cromwell assigns variables. The priority order is:
- use value from inputs json
- use default value set at workflow level
- use default value set at task level

So setting zero here means we use zero as the default but allow it to be overridden. This is the same way we're setting default values for other task parameters like preemptibles and cpu.

rexwangcc: I see I'm not familiar with this mechanism, thanks for the explanation here!

dshiga: There are a variety of use cases that I'm trying to keep in mind when setting the retries value. I think we should only have retries turned on in HCA production and scale testing. If someone is running this workflow outside of those use cases, they are either:
1. Testing out modifications - retries should be off because the changes are experimental with a good chance of failure, so retrying is a waste of time and resources
2. Reusing it for production outside HCA. Retries might make sense here but turning it off by default makes sense -- you have to opt into it deliberately if you want it on.

Similarly, someone might reuse individual tasks when testing out changes or building their own pipeline and I think it makes sense to have retries turned off by default in those situations, too.

The way we're setting the variables, we will only have it turned on when Lira launches the workflows (where it will set max retries as an input parameter), which will avoid having it turned on in most cases outside of production and scale testing.

rexwangcc: I haven't thought these use cases, this explanation totally makes sense to me. Thanks!

dshiga:
Hi @rexwangcc - I pushed a new commit to set the default at the workflow level to zero also. The motivation for that is to have mintegration tests run with retries turned off and also for anyone reusing our workflow to have retries turned off unless they deliberately choose to turn them on. Let me know what you think.

rexwangcc:
It does make sense to me given the generalizability beyond our own use case, as long as we could use the Lira config to control it. Thanks for making the change @dshiga !

library/tasks/HISAT2.wdl
```diff
@@ -120,6 +124,7 @@ task HISAT2RSEM {
     cpu: "(optional) the number of cpus to provision for this task"
     disk: "(optional) the amount of disk space (GB) to provision for this task"
     preemptible: "(optional) if non-zero, request a pre-emptible instance and allow for this number of preemptions before running the task on a non preemptible machine"
+    max_retries: "(optional) retry this number of times if task fails"
```
ambrosejcarr: It would be great to add some context from your conversation with rex above -- a caution that this is basically only to catch one-off errors that occur at production scale would be helpful for users.

library/tasks/RSEM.wdl
```diff
@@ -24,6 +25,7 @@ task RSEMExpression {
     cpu: "(optional) the number of cpus to provision for this task"
     disk: "(optional) the amount of disk space (GB) to provision for this task"
     preemptible: "(optional) if non-zero, request a pre-emptible instance and allow for this number of preemptions before running the task on a non preemptible machine"
+    max_retries: "(optional) retry this number of times if task fails"
```
ambrosejcarr: Same as above comment.

pipelines/smartseq2_single_sample/SmartSeq2SingleSample.wdl
```diff
@@ -55,14 +58,16 @@ workflow SmartSeq2SingleCell {
       fastq2 = fastq2,
       ref_name = hisat2_ref_name,
       sample_name = sample_name,
-      output_basename = quality_control_output_basename
+      output_basename = quality_control_output_basename,
```
ambrosejcarr: I like using trailing commas in lists because it cleans up diffs, but this is a style thing and I leave it up to you. 

(see how lines 79, 99, 107 aren't highlighted).

pipelines/smartseq2_single_sample/SmartSeq2SingleSample.wdl
```diff
@@ -44,6 +46,7 @@ workflow SmartSeq2SingleCell {
     output_name: "Output name, can include path"
     fastq1: "R1 in paired end reads"
     fastq2: "R2 in paired end reads"
+    max_retries: "Retry failed tasks this number of times"
```
ambrosejcarr: I would decide on one way to define this parameter meta and copy-paste it.

PR #118, opened 2018-06-27
[Not for merging] Mb debug optimus

ambrosejcarr:
- Puts Mehrtash's awesome debugging up for PR and adds a .py encoded version of the notebook for easier comments.


test/mb_debug_optimus.py
```diff
@@ -0,0 +1,667 @@
...
+import natsort
+import os
+from collections import namedtuple
+from collections import Counter
+import distance
+
+from typing import List, Set, Tuple
+# EPY: END code
+
+# EPY: START code
```
ambrosejcarr: Just for ease of reading, I'd move this down to where it's used.

test/mb_debug_optimus.py
```diff
@@ -0,0 +1,667 @@
...
+    primary_only_flag_and_summary_stats.process_difference(
+        op_flags, cr_flags, primary_only=True, joint_op='and')
+# EPY: END code
+
+# EPY: START code
+print(f'Number of Optimus reads with multiple primary alignments: {op_multiple_primary}, fraction = {op_multiple_primary/num_total_records:.4f}')
+print(f'Number of CellRanger reads with multiple primary alignments: {cr_multiple_primary}, fraction = {cr_multiple_primary/num_total_records:.4f}')
+print()
+
+print('Optimus (primary only):')
```
ambrosejcarr: In addition to your observations, I think it's interesting that Cell Ranger seems to make approximately twice as many reads as Optimus as duplicates. 

Related, I believe primary alignments are being marked randomly in cases where two reads have identical quality; a different random seed could explain this discrepancy in primary reads.

ambrosejcarr: The RC difference is a bit perplexing, since the alignments themselves are identical by name and num positions and were aligned to the same reference. Could that be a difference in STAR version? A bug somewhere?

ambrosejcarr: Looking at the numbers of primary, secondary, and duplicate alignments, cell ranger must mark some primary alignments as duplicates, which is not something that Picard does, I believe. 

Evidence: same number of secondary alignments between Optimus and Cell Ranger, but higher numbers of duplicates. 

If I had to guess, Cell Ranger might be marking  any replicated read as a duplicate, whereas Optimus does not mark primary reads that are duplicated as duplicates.

test/mb_debug_optimus.py
```diff
@@ -0,0 +1,667 @@
...
+
+# EPY: START markdown
+# **Summary**
+# 
+# - There are about ~ 70k identical raw reads in the synthetic data and they need to be removed
+# - For primary alignments with discordant flags, 47850 reads are marked as duplicate by both Optimus and CellRanger
+#   whereas 446777 reads (about half of all reads) are marked duplicate differently
+# - CellRanger marks more primary alignments as duplicate (356877 vs 185600), however, this does not necessary
+#   imply more molecules.
+# - It looks like `sctools` does not use the duplicate flag at all; `MarkDuplicates` only _marks_ duplicates. It does not remove them.
```
ambrosejcarr: That's true, but it _should_ use the duplicates flag. It'd be a faster way to construct the count matrix. I think we could just add up all the non-duplicate reads.

test/mb_debug_optimus.py
```diff
@@ -0,0 +1,667 @@
...
+# EPY: START code
+op_multiple_primary = 0
+cr_multiple_primary = 0
+
+for idx in discordant_flag_indices:
+    op_flags = [attr.flag for attr in op_master_rec_attr_list[idx]]
+    cr_flags = [attr.flag for attr in cr_master_rec_attr_list[idx]]
+    
+    # reads with more than one primary alignment (!)
+    # according to SAM specs, each read should have only one primary alignment
```
ambrosejcarr: I used to use `--outSAMprimaryFlag AllBestScore`. This was changed in the below commit. Given this change, I can't see why multiple primary alignments would be found in the data unless there are _duplicate reads_ (i.e. those with the same read names).  https://github.com/HumanCellAtlas/skylab/commit/e64c243c0b9c3180c0d086b0d5f0755ae6ad68ff#diff-c8197e3f1fec7111d0322013f2bd8707

test/mb_debug_optimus.py
```diff
@@ -0,0 +1,667 @@
...
+# EPY: END code
+
+# EPY: START markdown
+# **Summary**
+# 
+# - Close enough -- 95% concordance.
+# - I do not quite understand the difference; some thoughts:
+#   - Duplicate marking is immaterial for 10x since we only count unique (CB, GE, UB) tuples.
+#   - Multiple primary alignment is not an issue (unles CellRanger throws away reads with multiple primary alignments?)
+#   - Filtering by MQ could be an issue (Optimus does not filter, CellRanger does, but it manipulates the STAR bam out such that it is difficult to interpret their downstream filter MQ == 255)
```
ambrosejcarr: Looks like I never merged the multi-mapping changes. There is an intention to filter. 

https://github.com/HumanCellAtlas/sctools/commit/c81eb0a1f4634ca4116be3461e88a7c29f4b9f3e

Do the number of reads in Cell Ranger with MAPQ < 255 explain the difference in molecule number? If so, this could explain why we see more molecules, since we're currently not filtering.

ambrosejcarr:
This is no longer relevant.

PR #119, opened 2018-06-27
Dockstore SOP doc [not for merging yet]

mckinsel:
Description of process for publishing `skylab` pipelines in dockstore.

### DIFF
--- a/pipeline_release_process.md
+++ b/pipeline_release_process.md
@@ -30,3 +30,11 @@ Dockstore has some expectations about the organization of github repos that do n
     "SmartSeq2SingleCell.rrna_intervals": "http://storage.googleapis.com/hca-dcp-mint-test-data/reference/GRCh38_Gencode/gencode.v27.rRNA.interval_list",
     ...</pre></dd>
 </dl>
+
+Now we're ready to go to [Dockstore](https://dockstore.org/) and actually register the pipeline:
+
+- Go to dockstore.org and login in via github. You'll want to use the github account you use to commit to `skylab`
+- Click on "My Workflows", then "Register Workflow", then enter details about the entry point WDL. The repo name should be `HumanCellAtlas/skylab`, and the workflow path should be something like `/pipelines/smartseq2_single_sample/ss2_single_sample.wdl`.
+- Click "Register Workflow"
+- Select the `dockstore` branch from the dropdown menu and click "Publish"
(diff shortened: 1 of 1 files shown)


pipeline_release_process.md
```diff
@@ -0,0 +1,40 @@
...
+    ...</pre></dd>
+</dl>
+
+Now we're ready to go to [Dockstore](https://dockstore.org/) and actually register the pipeline:
+
+- Go to dockstore.org and login in via github. You'll want to use the github account you use to commit to `skylab`
+- Click on "My Workflows", then "Register Workflow", then enter details about the entry point WDL. The repo name should be `HumanCellAtlas/skylab`, and the workflow path should be something like `/pipelines/smartseq2_single_sample/ss2_single_sample.wdl`.
+- Click "Register Workflow"
+- Select the `dockstore` branch from the dropdown menu and click "Publish"
+- The workflow is now published, but it will be pointing to the most recently updated branch in `skylab`, so it's probably broken :(
```
mckinsel: Hey @garyluu @denis-yuen, this is where things kind of fall apart for me. I see that I'm supposed to select a "default" version in the versions tab. But there's also a version dropdown, and it's not clear how those interact or what the "Refresh" button does. Clicking it resets it to the wrong branch, and I never see anything show up in "Files".

garyluu: The versions dropdown is the version you're currently looking at which is what you'll see if the files tab, tools tab, DAG tab, etc.

https://docs.dockstore.org/docs/publisher-tutorials/workflows/ gives a brief description about the refresh button ("You can refresh and publish each repo that you identify as a real workflow in order to get, edit, and display additional information such as available versions.").  I personally only use it when I have a new version or if a file changed on GitHub since the last refresh.

For setting the default branch, it looks like you have the dockstore branch selected:
image

As long as the radio button for the default version has the branch selected, it should be fine.  I've created a ticket to investigate why it's bouncing back to a different version after user actions: 
https://github.com/ga4gh/dockstore/issues/1553

mckinsel: Thanks Gary. I do see all the correct files now. Is it possible that it takes a little time for the files to update?

garyluu: It shouldn't, my guess is that a different branch version was displayed

nikellepetrillo:
@ekiernan @kbergin Do we still want to add this documentation?

kbergin:
We do, but we'll update it for the current state of the world in WARP.

PR #120, opened 2018-07-16
run SS2 in batch

jishuxu:
In this PR
1) a new workflow to run ss2 samples by plate 
2) add new functions to aggregate Picard QC results. 
3) add a function to generate a data matrix per plate

This PR is not ready to merged yet but it will give us some idea what we will expect if we start to use plate bundle instead of single cell bundle


pipelines/smartseq2_by_plate/AggregatePicardMetrics.py
```diff
@@ -0,0 +1,45 @@
...
+    
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-f",
+        "--file_names",
+        dest="file_names",
+        required=True,
+        help=
+        "a list of files to be parsed out."
```
mckinsel: You can use `nargs='+'` to allow multiple arguments that get turned directly into a list.

pipelines/smartseq2_by_plate/AggregatePicardMetrics.py
```diff
@@ -0,0 +1,45 @@
...
+    piepline output picard QC metrics at sinle cell/sample level.
+    This functin is called to merge/aggregate QC metrics by metrics type and then merge multiple QC measurement 
+    into single matrix file. In this file, column is sample/cell and row is QC metrics
+    :param files: metric files from pipeline outputs
+    :param met_name: metrics name with workflow name and subworkflow name as prefix. such as 'run_pipelines.RunStarPipeline.alignment_summary_metrics'
+    """
+    # initial output
+    print(files)
+    filenames = files.split(',')
+    for kk in range(0, len(filenames)):
```
mckinsel: Think you can drop the `0`, `range`s start at 0 by default.

pipelines/smartseq2_by_plate/AggregatePicardMetrics.py
```diff
@@ -0,0 +1,45 @@
+import json
+import argparse
+import pandas as pd
+
+def AggregatePicardMetric(files, output_name):
```
mckinsel: Generally python function names should be lowercase, underscore-separated.

pipelines/smartseq2_by_plate/SmartSeq2Plate.wdl
```diff
@@ -0,0 +1,133 @@
+import "SmartSeq2SingleSample.wdl" as single_cell_run
+
+task AggregateDataMatrix{
+  Array[File] filename_array
+  String col_name
+  String output_name
+  String docker = "gcr.io/broad-dsde-mint-dev/benchmarking-tools:0.0.1"
+  command {
+    set -e
+    gsutil cp gs://broad-dsde-mint-dev-teststorage/pipeline_testing_scripts/MergeDataMatrix.py ./
```
mckinsel: This seems a little brittle. I think someone would be surprised that deleting or moving a script in a google bucket would break other WDLs.

pipelines/smartseq2_by_plate/SmartSeq2Plate.wdl
```diff
@@ -0,0 +1,133 @@
...
+  File hisat2_ref_trans_index
+  File rsem_ref_index
+  # ref index name
+  File hisat2_ref_name
+  File hisat2_ref_trans_name
+  # samples
+  String stranded
+  String sra_dir
+  Array[String] out_types = ["est_counts","tpm"]  
+  Array[String] sraIDs
```
dshiga: This workflow should expect an Array[Object] cells, which will be provided by the adapter workflow. For testing this workflow, you can specify the array in your inputs.json like this:
```
  "RunSmartSeq2ByPlate.cells": [
    {
      "sample_name": "sample_1",
      "fastq_1": "gs://hca-dcp-mint-test-data/smartseq2_single_sample/foo/sample_1_read1.fastq.gz",
      "fastq_2": "gs://hca-dcp-mint-test-data/smartseq2_single_sample/foo/sample_1_read2.fastq.gz"
    },
    {
      "sample_name": "sample_2",
      "fastq_1": "gs://hca-dcp-mint-test-data/smartseq2_single_sample/foo/sample_2_read1.fastq.gz",
      "fastq_2": "gs://hca-dcp-mint-test-data/smartseq2_single_sample/foo/sample_2_read2.fastq.gz"
    }
  ]
```
Then you can access the properties like
```
scatter(cell in cells) {
  call single_cell_run.SmartSeq2SingleCell as sc {
    fastq1 = cell.fastq_1,
    fastq2 = cell.fastq_2,
    sample_name = cell.sample_name,
    ...(etc)...
  }
} 
```

pipelines/smartseq2_by_plate/SmartSeq2Plate.wdl
```diff
@@ -0,0 +1,133 @@
...
+        sample_name = sraIDs[idx],
+        output_name = sraIDs[idx]
+   }
+   call AggregatePicardMetricsByCell {
+      input:
+        metric_files = [sc.alignment_summary_metrics,sc.rna_metrics, sc.dedup_metrics,sc.insert_size_metrics],
+        output_name = sraIDs[idx]
+      } 
+  }
+  scatter(i in range(length(out_types))){
```
dshiga: You can just say:
```
scatter(out_type in out_types) {
  call AggregateDataMatrix as AggregateGene {
    input:
      filename_array = sc.rsem_gene_results,
      col_name = out_type,
      output_name ="gene_" + out_type + ".csv"
  }
  ...(etc)...
```

pipelines/smartseq2_by_plate/SmartSeq2Plate.wdl
mckinsel: Err, the `Array[Object]` would hide the input files in the object, which is untyped. This isn't going to be good for portability to any system that doesn't handle gs url strings the way cromwell does.

dshiga: Hmm, if we can't use Objects, then I think the only other viable option is to use 3 parallel arrays:
```
{
  "RunSmartSeq2ByPlate.sample_names": ["sample_1", "sample_2"],
  "RunSmartSeq2ByPlate.fastq_1s": ["gs://foo/sample_1_read1.fastq.gz", "gs://foo/sample_2_read1.fastq.gz"],
  "RunSmartSeq2ByPlate.fastq_2s": ["gs://foo/sample_1_read2.fastq.gz", "gs://foo/sample_2_read2.fastq.gz"]
}
```

And use like:
```
scatter(idx in range(length(sample_names))) {
  call single_cell_run.SmartSeq2SingleCell as sc {
    fastq1 = fastq_1s[idx],
    fastq2 = fastq_2s[idx],
    sample_name = sample_names[idx],
    ...(etc)...
  }
} 
```

mckinsel: Yeah, I think that's probably what we'd have to do. I'm pretty sure that's how optimus is handling a similar concern.

Also worth taking a look at this PR: https://github.com/openwdl/wdl/pull/228

ambrosejcarr: `Struct` was tossed around as a possible fix for the low-expressivity of `Object`, could that help at all here? I recall being hopeful that it could solve the Optimus problem that you point out is similar but I haven't revisited it yet.

mckinsel: I think it would help a lot, though I'm not sure of its implementation status.

dshiga: The problem is there is no way to dynamically create an array of structs right now. The adapter wdl is going to parse the metadata to get the lists of fastqs and sample names but there is currently no way for us to then get that info into an array of structs. The various write_json / read_json and similar functions have limitations that prevent this currently. I do hope in the longer term when those limits go away that we can refactor to use structs though.

cjllanwarne: @dshiga what are those limitations? Are they WDL limitations or Cromwell limitations that we should fix?

cjllanwarne: If you use object the values *will* come in as `String` but the proposed answer by @dshiga will still work because they will be coerced into `File` at the point when the individual fields becomes inputs to the call:
```wdl
scatter(cell in cells) {
  call single_cell_run.SmartSeq2SingleCell as sc {
    fastq1 = cell.fastq_1,
    fastq2 = cell.fastq_2,
    sample_name = cell.sample_name,
    ...(etc)...
  }
} 
```

cjllanwarne: Despite my above comment, I still think `struct` should be the preferred answer here (pending addressing the limitation), though it does also imply a WDL version upgrade to 1.0 (right now this WDL is draft-2)

mckinsel: @cjllanwarne, the object solution is the "a wild File appears!" behavior that I'd really prefer we avoid in HCA WDLs. It doesn't work very well outside of cromwell/google.

pipelines/smartseq2_by_plate/SmartSeq2Plate.wdl
cjllanwarne: I agree - 
I'm curious why was a hard-coded String instead of a `File` input? It feels like something we should try to address in the spec if WDL encourages (or forces!) hard-coded file paths

pipelines/smartseq2_by_plate/SmartSeq2Plate.wdl
dshiga: @cjllanwarne the limitation is that there is no way to dynamically create an array of structs. The way our wdls are set up, we have an outer wdl that would need to create the array of structs dynamically by making an http request to the blue box and parsing the response. Then it would need to pass the array as an input to the inner workflow. But none of the read_json or similar methods appear capable of creating an array of structs. I think the only way to create an array of structs is by specifying it literally in the input json to the workflow, but that won't work in our situation.

pipelines/smartseq2_by_plate/AggregateMetrics.py
```diff
@@ -0,0 +1,99 @@
+from crimson import picard
+import json
+import argparse
+import os
+import pandas as pd
+
+
+def AggregatePicardMetricsRow(filenames, output_name):
```
dshiga: This file feels like it belongs in the sctools repo. Also, can all of this be done using the Picard jar, or are there limitations that requires us to write our own code for this?

pipelines/smartseq2_by_plate/SmartSeq2Plate.wdl
dshiga: In a discussion offline, @cjllanwarne suggested a good workaround, where we would define two levels of structs, something like:
```
struct Cell {
  File fastq_1
  File fastq_2
  String sample_id
}

struct CellArray {
  Array[Cell] cells
}
```

We could populate that outer struct using read_json.

However, this seems a bit exotic, so I am wondering whether this, too, would be a portability problem. What do you think @mckinsel ?

PR #124, opened 2018-08-13
Add cellranger.wdl

samanehsan:
Add a WDL that runs the cellranger count command, and a Dockerfile that builds cellranger from the [source code](https://github.com/10XGenomics/cellranger) and installs all of its requirements. 

Addresses ticket: https://github.com/HumanCellAtlas/secondary-analysis/issues/245

- [ ] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/docker/cellranger/Dockerfile
+++ b/docker/cellranger/Dockerfile
@@ -23,7 +23,7 @@ ENV PATH $PATH:/Python-${pyversion}:/usr/lib/go-1.9/bin:/root/.cargo/bin
 RUN rustup install 1.19.0 && rustup default 1.19.0
 
 ## CellRanger 2.1.1
-RUN apt-get -y install git && git clone https://github.com/samanehsan/cellranger.git
+RUN apt-get -y install git && git clone https://github.com/10XGenomics/cellranger.git
 WORKDIR cellranger
 RUN make
 WORKDIR /root


docker/cellranger/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+`docker build .`
+
+## Interactively work with image
+
+`docker run -it <image> /bin/bash`
+
+## Verify with the following command:
+
+`docker run <image>`
```
ambrosejcarr: I'd like some kind of basic test that shows cellranger working. Ideally that would be on testing data, but if you're stretched I'd take an invocation of the binary that prints help or something. :-)

samanehsan: How about this existing test @ambrosejcarr? https://github.com/HumanCellAtlas/skylab/blob/master/test/trigger_cellranger_test.sh

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,75 @@
...
+        File raw_matrix = "${sampleId}/outs/raw_gene_bc_matrices/${reference}/matrix.mtx"
+        File mol_info_h5 = "${sampleId}/outs/molecule_info.h5"
+    }
+
+    runtime {
+        docker: "quay.io/humancellatlas/secondary-analysis-cellranger"
+        memory: "416 GB"
+        bootDiskSizeGb: 12
+        disks: "local-disk ${diskSpace} HDD"
+        cpu: 64
```
samanehsan: Do we want to make this `preemptable`? I took it out while testing because the wdl was taking twice as long to run.

dshiga: I think it's fine as is for now. We can make it preemptible later if we want to optimize for cost.

docker/cellranger/Dockerfile
```diff
@@ -0,0 +1,85 @@
+FROM ubuntu:16.04
+
+## Install Variables
+ENV pyversion 2.7.13
+
+## Dependencies python 2.7.13
+## rust 1.19.0
```
dshiga: Can these commented out lines be removed?

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,92 @@
...
+        File raw_gene_h5 = "${sample_id}/outs/raw_gene_bc_matrices_h5.h5"
+        File raw_barcodes = "${sample_id}/outs/raw_gene_bc_matrices/${reference}/barcodes.tsv"
+        File raw_genes = "${sample_id}/outs/raw_gene_bc_matrices/${reference}/genes.tsv"
+        File raw_matrix = "${sample_id}/outs/raw_gene_bc_matrices/${reference}/matrix.mtx"
+        File mol_info_h5 = "${sample_id}/outs/molecule_info.h5"
+    }
+
+    runtime {
+        docker: "quay.io/humancellatlas/secondary-analysis-cellranger"
+        memory: "416 GB"
```
dshiga: Maybe we should expose these as parameters with the existing values as defaults. We have small test datasets that could run on a much smaller machine.

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,92 @@
...
+    String reference
+    File transcriptome_tar_gz
+    Int? expect_cells
+    String disk_space
+
+    command {
+        set -e
+        mkdir transcriptome_dir
+        tar xf ${transcriptome_tar_gz} -C transcriptome_dir --strip-components 1
+        ln -s /usr/bin/python3 python
```
dshiga: Should this be done in the Dockerfile instead?

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,92 @@
...
+        export PATH=$PATH:.
+        python <<CODE
```
dshiga: I know this is copy pasted from elsewhere, but it would be helpful to have some newlines in the command block and maybe some comment lines to explain what's happening.

docker/cellranger/README.md
dshiga: We should provide an example inputs json file next to the cellranger wdl file as part of this PR. We should probably also set up a some test automation but that could happen in a later PR.

samanehsan: For local testing, I was using a `transcriptome_tar_gz` file in one of our team gcloud buckets. But I believe it was originally downloaded from cellranger and required agreeing to their software license, so we would need to provide our own reference in the example inputs if my understanding is correct here.

dshiga: Hmm, okay, @ambrosejcarr or @mckinsel, do you know if we have our own transcriptome tar file that we could use or is this something that has to come from 10x?

pipelines/cellranger/cellranger.wdl
samanehsan: I'm wondering if we can actually remove those lines, since Cellranger lists python 2.7.13 under its [build dependencies](https://github.com/10XGenomics/cellranger#build-dependencies) and that is the python version installed in the Dockerfile. Tagging @TimothyTickle for some clarification here, since he originally put this together!

docker/cellranger/Dockerfile
```diff
@@ -0,0 +1,80 @@
+FROM ubuntu:16.04
+
+## Install Variables
+ENV pyversion 2.7.13
+
+## Python
+RUN apt-get update && apt-get -y install build-essential checkinstall libreadline-gplv2-dev libncursesw5-dev libssl-dev libsqlite3-dev tk-dev libc6-dev libbz2-dev wget make && wget https://www.python.org/ftp/python/${pyversion}/Python-${pyversion}.tgz && tar -xvf Python-${pyversion}.tgz
```
mckinsel: Does it really fail with the 2.7.12 that comes with ubuntu 16.04?

docker/cellranger/Dockerfile
```diff
@@ -0,0 +1,80 @@
...
+RUN apt-get -y install curl && curl https://sh.rustup.rs -sSf | sh -s -- -y
+
+## Set Path
+ENV PATH $PATH:/Python-${pyversion}:/usr/lib/go-1.9/bin:/root/.cargo/bin
+
+## RUST finish
+RUN rustup install 1.19.0 && rustup default 1.19.0
+
+## CellRanger 2.1.1
+RUN apt-get -y install git && git clone https://github.com/samanehsan/cellranger.git
```
mckinsel: Can this point to the 10x repo or maybe a fork in HCA?

docker/cellranger/README.md
mckinsel: You can make your own using `cellranger mkref`, which I'm pretty sure Ambrose has done a few times. I thought they had a small one available on their website, but you now have to agree to their privacy policy, which I do not.

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,112 @@
...
+
+    # Runtime attributes
+    String memory = "416 GB"
+    Int boot_disk_size_gb = 12
+    String disk_space = "250"
+    Int cpu = 64
+
+    parameter_meta {
+        sample_id: "Name of sample to run CellRanger count on"
+        fastqs: "Array of fastq files for running decoupled"
```
mckinsel: I'm not sure I understand what this means. What's "decoupled"?

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,112 @@
...
+        reference_name: "Reference name for count"
+        transcriptome_tar_gz: "Cellranger transcriptome reference"
+        expect_cells: "Expected number of recovered cells (defaults to 3000)"
+        memory: "The minimum amount of RAM to use"
```
mckinsel: This sounds a little like you're telling cellranger to use this much memory.

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,112 @@
...
+    Int? expect_cells
+    String memory
+    Int boot_disk_size_gb
+    String disk_space
+    Int cpu
+
+    command {
+        set -e
+        mkdir transcriptome_dir
+        tar xf ${transcriptome_tar_gz} -C transcriptome_dir --strip-components 1
```
mckinsel: Extracting with `--no-same-owner` seems to avoid bugs in some environments.

pipelines/cellranger/cellranger.wdl
mckinsel: Is this just for the python heredoc?

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,112 @@
...
+        memory: "The minimum amount of RAM to use"
+        boot_disk_size_gb: "Size of disk (GB) where the docker image is booted"
+        disk_space: "Amount of disk space (GB) to allocate to the VM"
+        cpu: "The minimum number of cores to use"
+    }
+
+    call CellRanger {
+        input:
+        sample_id = sample_id,
+        fastqs = fastqs,
```
mckinsel: I think we're inheriting the restrictions on cellranger fastq naming and organization, right? We might want to make a little note somewhere.

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,112 @@
...
+        export PATH=$PATH:.
+
+        python <<CODE
+        import os
+        from subprocess import call
+
+        # Get paths to fastq directories
+        dirs = dict()
+        for f in ["${sep='","' fastqs}"]:
+            dirs.setdefault(os.path.dirname(f), True)
```
mckinsel: You really just want a `set` here I think.

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,112 @@
...
+        call_args.append('count')
+        call_args.append('--jobmode=local')
+        call_args.append('--transcriptome=transcriptome_dir')
+        call_args.append('--id=' + '${sample_id}')
+        call_args.append('--fastqs=' + ','.join(list(dirs.keys())))
+        call_args.append('--nosecondary')
+        call_args.append('--disable-ui')
+
+        expect_cells = '${expect_cells}'
+        if expect_cells is not '':
```
mckinsel: `if expect_cells:`

pipelines/cellranger/cellranger.wdl
```diff
@@ -0,0 +1,112 @@
...
+            call_args.append('--expect-cells=' + str(expect_cells))
+        call(call_args)
```
mckinsel: We probably want to `check_call` here.

docker/cellranger/Dockerfile
samanehsan: Good catch! I forgot to change that!

pipelines/cellranger/cellranger.wdl
samanehsan: I found this definition in another version of the cellranger wdl, but I'll take it out since it doesn't seem to mean much 😅

docker/cellranger/Dockerfile
samanehsan: Hmm I haven't tried! Cellranger lists Python 2.7.13 under their build dependencies: https://github.com/10XGenomics/cellranger#build-dependencies but I'm not sure what relies on that particular version.

pipelines/cellranger/cellranger.wdl
samanehsan: This makes the python code in the `command` block use python3, but nothing in the code requires that so I'll remove this.

docker/cellranger/README.md
samanehsan: Ok! For now I put the cellranger reference in one of our private mint team gcloud buckets, and we can replace that in the example inputs when we have a reference of our own, which I think can be a separate ticket.

samanehsan:
Thanks for the comments @dshiga and @mckinsel! Let me know if these changes look good to you!

mckinsel:
Yep looks good. Thanks!

PR #125, opened 2018-08-14
Use sctools bam sorting instead of samtools

dshiga:
Sort bams using new sctools bam sorting.

See https://app.zenhub.com/workspace/o/humancellatlas/secondary-analysis/issues/170

- [x] Added or updated tests
- [ ] Updated documentation

See related: https://github.com/HumanCellAtlas/sctools/pull/46

### DIFF
--- a/library/tasks/TagSortBam.wdl
+++ b/library/tasks/TagSortBam.wdl
@@ -2,7 +2,7 @@ task CellSortBam {
   File bam_input
 
   # runtime values
-  String docker = "quay.io/humancellatlas/secondary-analysis-samtools:v0.2.2-1.6"
+  String docker = "quay.io/humancellatlas/secondary-analysis-sctools:0.2.0"
   Int machine_mem_mb = 7500
   Int cpu = 2
   Int disk = ceil(size(bam_input, "G") * 8)
@@ -24,7 +24,7 @@ task CellSortBam {
   command {
     set -e
 
-    samtools sort -t GE -t UB -t CB -o cell-sorted.bam "${bam_input}"
+    TagSortBam -i ${bam_input} -o cell-sorted.bam -t CB UB GE
   }
   
   runtime {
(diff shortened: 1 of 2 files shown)


dshiga:
retest this please

dshiga:
retest this please

library/tasks/TagSortBam.wdl
```diff
@@ -24,7 +24,7 @@ task CellSortBam {
   command {
     set -e
 
-    samtools sort -t GE -t UB -t CB -o cell-sorted.bam "${bam_input}"
+    TagSortBam -i ${bam_input} -o cell-sorted.bam -t GE -t UB -t CB
```
samanehsan: I'm slightly confused by the order of tags here, since the description for the `CellSortBam` task is `"Sort bam_input by cell, then molecule, then gene."`

dshiga: Good catch - I've changed the order to match the description. Depending on the tool being used, there are cases where you would need to list things in the reverse order to the actual sort order you want. But that is definitely not the case for the sctools sorting tool.

samanehsan: That's good to know! Thanks!

samanehsan: It looks like there are a few tests that may need to be updated as well!

dshiga:
retest this please

samanehsan:
@dshiga do we know why the optimus tests are failing?

dshiga:
retest this please

dshiga:
Yeah, there was a bug to do with missing tags bubbling up as `None` when we need it to be an empty string. Fixed it.

library/tasks/TagSortBam.wdl
```diff
@@ -2,7 +2,7 @@ task CellSortBam {
   File bam_input
 
   # runtime values
-  String docker = "quay.io/humancellatlas/secondary-analysis-samtools:v0.2.2-1.6"
+  String docker = "quay.io/humancellatlas/secondary-analysis-sctools:ds-multi-tag-sort-170"
```
ambrosejcarr: You'll want to update this to a tagged version of master prior to merge, right?

dshiga:
retest this please

library/tasks/TagSortBam.wdl
dshiga: Definitely, will do

dshiga:
retest this please

PR #127, opened 2018-08-31
Add version parameter to all of the analysis pipelines except the deprecated 10x.

rexwangcc:
I expect this PR should wait for @jishuxu's PR about outputs changes to be merged in so we can tag them together as `smartseq2_v2.0.0`

- [ ] Added or updated tests
- [ ] Updated documentation

### DIFF
--- a/pipelines/cellranger/cellranger.wdl
+++ b/pipelines/cellranger/cellranger.wdl
@@ -2,6 +2,8 @@ workflow CellRanger {
     meta {
         description: "Analyze 3' single-cell RNA-seq data using the 10X Genomics Cellranger pipeline."
     }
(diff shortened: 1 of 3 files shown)


samanehsan:
I just saw that the circleci tests are failing, could you take a look at those in case they need to be updated as well?

rexwangcc:
@samanehsan Yeah, I have no idea why the portability tests failed, consulting @mckinsel now, the worst case would be that adding a version to the pipelines is destroying the portability :(

rexwangcc:
The latest failed Mintegration test was because that the test suite itself is not updated to match the master branch of Lira.

rexwangcc:
retest this please

PR #129, opened 2018-09-07
Add matrix format output spec

mckinsel:
For review


docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+stored within a single zarr group called `expression_matrix`.
+
+- expression - The expression values themselves. Rows correspond to cells and
+  columns to genes. These must represent raw counts. Any normalization is
+  deferred, but metadata that enables normalization should be included in the
+  metadata arrays defined below.
+  - shape: `(n_cells, n_genes)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(10000, n_genes)`
+  - dtype: `int32`
```
brianraymor: `uint32` for raw counts.

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+
+- gene_metadata - Numeric metadata values associated with each gene. Like
+  `cell_metadata` these can vary by experiment type, but care should be taken
+  to maintain consistency. 
+  - shape: `(n_gene_metadata_values, n_genes)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(n_gene_metadata_values, n_genes)`
+  - dtype: `float32`
+
+- cell_metadata_name - Field names for `cell_metadata`. These must be unique.
```
brianraymor: For consistency with gene_id - _Elements in this array must be unique._ ?

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+  - shape: `(n_cell_metadata_values,)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(n_cell_metadata_values,)`
+  - dtype: `U40`
+
+- gene_metadata_name - Field names for `gene_metadata`. These must be unique.
```
brianraymor: For consistency with gene_id - _Elements in this array must be unique._ ?

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+  - shape: `(n_gene_metadata_values,)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(n_gene_metadata_values,)`
+  - dtype: `U40`
+
+# Example Bundle Structure
+
+To illustrate how the zarr directore store manifests in an HCA DCP bundle,
```
brianraymor: directore->directory

docs/matrix_format_spec.md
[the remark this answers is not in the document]
mckinsel: Okay thanks, I can change to `float32`. Just so I understand, what do fractional raw counts represent?

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+  deferred, but metadata that enables normalization should be included in the
+  metadata arrays defined below.
+  - shape: `(n_cells, n_genes)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(10000, n_genes)`
+  - dtype: `int32`
+
+- cell_id - Identifiers for each cell. The means of identifying a cell varies
+  by experiment type. For example, in droplet data this is a barcode; in
+  well-based assays it is a well identifier. The `ith` element of this array
```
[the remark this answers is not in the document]
mckinsel: Thanks Jishu, do you know now that might work when we have multiple cells in one SS2 bundle? It does seem like you'd need a way to refer to a cell.

docs/matrix_format_spec.md
mckinsel: :+1:

docs/matrix_format_spec.md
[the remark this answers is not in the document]
dshiga: Looks like the metadata schema for ss2 has the ability to capture well names and/or well row/column. I do recall there were some concerns about making these fields required so I believe they're optional right now. Is that correct @malloryfreeberg?
https://github.com/HumanCellAtlas/metadata-schema/blob/master/json_schema/module/process/sequencing/smartseq2.json#L24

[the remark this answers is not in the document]
malloryfreeberg: `plate_id` is required for plate-based sequencing (e.g. SS2). `well_id` (An ID or name for the well. Should be unique for the plate) is _not_ required, but as you said, there was a lot of discussion about this. Each cell _will_ have a unique ID in the HCA metadata.

@dshiga the link you indicated is to the production schema, but the most recent version of the `plate_based_sequencing.json` schema (formerly `smartseq2.json`) has only been promoted up to staging. The recent use for `well_id` can be seen here: https://github.com/HumanCellAtlas/metadata-schema/blob/staging/json_schema/module/process/sequencing/plate_based_sequencing.json#L29

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+Expression values and metadata are stored in seven arrays. These arrays are
+stored within a single zarr group called `expression_matrix`.
+
+- expression - The expression values themselves. Rows correspond to cells and
+  columns to genes. These must represent raw counts. Any normalization is
+  deferred, but metadata that enables normalization should be included in the
+  metadata arrays defined below.
+  - shape: `(n_cells, n_genes)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(10000, n_genes)`
```
samanvp: I think we should emphasize the optimal chunk size depends on many factors. First, size of a chunk depends on:
* dtype
* n_genes
* sparsity of the matrix
* compression algorithm
* and the number of cells in each chunk (here 10000)

The last item works as a knob to adjust the chunk size. The optimal chunk size depends on the storage option. Having too many small chunks means more frequent disk requests while having a few larger chunks means more [read delay](https://medium.com/@duhroach/gcs-read-performance-of-large-files-bd53cfca4410).

In my tests on GCS I was aiming for 2-4 MB chunk sizes, I am not sure what would be a good option for AWS. [This](https://github.com/zarr-developers/zarr/issues/86) might be relevant.

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+  - dtype: `int32`
+
+- cell_id - Identifiers for each cell. The means of identifying a cell varies
+  by experiment type. For example, in droplet data this is a barcode; in
+  well-based assays it is a well identifier. The `ith` element of this array
+  identifies the `ith` row of the `expression` array. Elements in this array
+  must be unique.
+  - shape: `(n_cells,)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(10000,)`
```
samanvp: I am not sure what is a better practice here, to chunk or not to chunk.
Since this is a small matrix, we could load it once into memory. However, I am not sure if it's better to have consistent chunks along each axis?

docs/matrix_format_spec.md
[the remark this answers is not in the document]
ambrosejcarr: > SS2 outputs are decimal values not integer even for raw counts

Could you explain this Jishu? This isn't consistent with my understanding. I suspect maybe we have different ideas about what raw counts are in this case. Do you mean estimated counts from RSEM? Gene counts would still be integers, right?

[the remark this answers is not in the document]
TimothyTickle: I think this is the estimated gene counts after the EM algorithm of RSEM. I think this makes sense for use to deliver.

docs/matrix_format_spec.md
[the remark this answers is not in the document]
TimothyTickle: Note cells may be unique to the run but 10X barcodes are not explicitly unique between runs. In the delivery of the matrix to the user, there needs to be barcode + run or some extra information added to the id to assure uniqueness.

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+- gene_id - Identifiers for each gene. The `ith` element of this array
+  identifies the gene at the `jth` column of the `expression` array. The
+  `gene_id` array must be identical for all expression data produced by HCA
+  secondary analysis pipelines. Elements in this array must be unique.
+  - shape: `(n_genes,)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(n_genes,)`
+  - dtype: `U40`
+
+- cell_metadata - Numeric metadata values associated with each cell. These
```
TimothyTickle: Does this mean all cell metadata (and gene metadata) must be unique? I am not sure that assumption holds or will hold.

docs/matrix_format_spec.md
mckinsel: :+1:

docs/matrix_format_spec.md
mckinsel: 👍

docs/matrix_format_spec.md
mckinsel: So it turns out this actually needs to be a `float32`. For some experiment types, the "raw count" is a sum of probabilities.

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,101 @@
...
+  well-based assays it is a well identifier. The `ith` element of this array
+  identifies the `ith` row of the `expression` array. Elements in this array
+  must be unique.
+  - shape: `(n_cells,)`
+  - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
+  - chunk shape: `(10000,)`
+  - dtype: `U40`
+
+- gene_id - Identifiers for each gene. The `ith` element of this array
+  identifies the gene at the `jth` column of the `expression` array. The
```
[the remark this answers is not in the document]
mckinsel: :+1:

docs/matrix_format_spec.md
mckinsel: Hey thanks Saman. Yeah you're right that there are a number of considerations for determining the optimal chunk size. I also think there are two different places where we have to determine that: 1) in the HCA's data storage system and 2) in the matrices that are delivered to users. Here we're talking pretty much only about (1), and there I'm mostly optimizing for what an AWS lambda can hold in its maximum 3GB of memory and 5 minute running time. It's a little hand-wavey, but 10000 cells x 30000 genes seems to work.

For (2) we actually have a lot more flexibility and diversity in use cases. The size of the chunks I'm producing with this spec are ~40MB, but we can do some more testing around what we can deliver to users.

docs/matrix_format_spec.md
mckinsel: Yeah it's very small. The idea behind the chunking here is to better support parallel writing.

docs/matrix_format_spec.md
mckinsel: I think we just want to make sure that field names aren't duplicated anywhere within in single array.

samanehsan:
@mckinsel the integration tests are probably failing here because this branch is not up to date with the master branch.

kbergin:
@mckinsel Can this be merged in?

kbergin:
@chrisfarnham

docs/matrix_format_spec.md
```diff
@@ -0,0 +1,144 @@
...
+tabular metadata using the zarr version 2 directory store format. The format is
+specified more precisely
+[here](https://zarr.readthedocs.io/en/stable/spec/v2.html), but in general zarr
+stores contain groups and arrays. Groups can contain other groups and arrays.
+Arrays are stored chunked and compressed. The file names and directory
+structure of the zarr store convey the group and chunk structure.
+
+# Groups and Arrays
+
+Expression values and metadata are stored in seven arrays. These arrays are
```
samanehsan: Should this say that the `Expression values and metadata are stored in eleven arrays` instead of seven?

mckinsel: Yep, thanks

PR #130, opened 2018-09-24
Add a build reference accessory workflow for cellranger

rexwangcc:
We ended up with using the Ensembl instead of Gencode data for building the reference, as the 10x documentation showed [here](https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/advanced/references), to reproduce the promising result we got from running the cellranger pipeline against their official 1.2.0 reference.

The reference built by this workflow is uploaded to the public bucket now. The testing workflow runs for this PR are:
- https://job-manager.mint-dev.broadinstitute.org/jobs/3a7dd7cf-6394-408a-8a79-9d14a792c642
- https://job-manager.mint-dev.broadinstitute.org/jobs/bb2d50b2-37ba-4521-8bd9-5486bfe4ca8a

---
- [ ] Added or updated tests
- [ ] Updated documentation


library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
+task BuildCellRangerReference {
+  String ref_fasta
+  String gtf_file
+  String ref_name
+
+  String docker = "quay.io/humancellatlas/secondary-analysis-cellranger"
+  String memory = "416 GB"
```
dshiga: Do we definitely need a machine this big?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
...
+                     --attribute=gene_biotype:TR_J_pseudogene \
+                     --attribute=gene_biotype:TR_C_gene
+
+
+    cellranger mkref --genome=GRCh38 \
+                     --fasta="$ref_fasta_filename" \
+                     --genes="filtered.$gtf_filename" \
+                     --nthreads=$(getconf _NPROCESSORS_ONLN)
+
+    tar cvf ${ref_name}.tar GRCh38/
```
dshiga: We could save some space and file transfer time by gzipping it. Is there a reason why we wouldn't want to?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
+task BuildCellRangerReference {
+  String ref_fasta
+  String gtf_file
+  String ref_name
+
+  String docker = "quay.io/humancellatlas/secondary-analysis-cellranger"
```
dshiga: Should we add a v1.0.0 or some other tag to this repo and then reference that here (or perhaps pass in the tag from the inputs json)? Otherwise runs of this wdl may not be easily repeatable.

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
...
+  }
+  output {
+    File cellRangerRef = "${ref_name}.tar"
+  }
+}
+
+workflow BuildCellRangerRef {
+  String fasta
+  String gtf
+  String ref_name
```
dshiga: Could you please add some meta and parameter meta for the workflow and the task?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
+task BuildCellRangerReference {
+  String ref_fasta
+  String gtf_file
+  String ref_name
+
+  String docker = "quay.io/humancellatlas/secondary-analysis-cellranger"
+  String memory = "416 GB"
+  Int boot_disk_size_gb = 12
+  String disk_space = "250"
```
dshiga: Would be better to use Int here and then use below like: `"local-disk ${disk} HDD"`.  Similarly for memory.

rexwangcc: will do!

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
rexwangcc: will do!

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
rexwangcc: That makes sense to me! Being that said, do you think I'd better also change the cellranger wdl we have for the same reason? https://github.com/HumanCellAtlas/skylab/blob/ffd17882c831da377dce3ab471c7a44683dd00f6/pipelines/cellranger/cellranger.wdl#L110

dshiga: Yes, that sounds like a good idea! Any tag is better than nothing.

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
rexwangcc: No, there are no special reasons for doing this, except:
- I thought using tar.gz will cause the cellranger run slower
- I copied some code from existing build reference wdls, which are using tars
I think we can use `tar cvzf` here!

dshiga: Hmm, yeah it will make cellranger run a little slower if it has to gunzip. (Copy time will be slightly faster but it seems it's outweighed by decompression time based on an old discussion in slack.) Maybe it's better to not gzip for now and we can add it later if we need to.

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
rexwangcc: https://cloud.google.com/compute/docs/machine-types is a good info to know! So let me use 60 GB for a try. When it comes to CPUs, I used `getconf _NPROCESSORS_ONLN` which will fetch the number of available CPU cores in the runtime.

dshiga: Oh, I see, the original settings result in getting n1-highmem-64. However, since the documentation says we don't need so much memory, it might make sense to use n1-highcpu-64 instead, which we should get if we ask for 64 cpus and ~50 GB of memory.

jishuxu: we asked for 50G when we built Star reference for  SS2. it should be adequate amount. The lower MEM might work as well.

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
jishuxu: Yea, we ran into the same problem while we were working with SS2.  `gunzip`  definitely takes some time

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
...
+  Int cpu = 64
+
+  command {
+    set -e
+
+    wget ${ref_fasta}
+    ref_fasta_tar_filename=$(echo ${ref_fasta} | sed 's:.*/::')
+    gunzip $ref_fasta_tar_filename
```
mckinsel: Few little things:
- I think this isn't a tar archive, so probably want to remove tar from the variable name.
- You can download and decompress in parallel
- I think `basename` is a more straightforward way to get the local name

So something like this:
```bash
ref_fasta_filename="$(basename "${ref_fasta}")"
ref_fasta_filename="${ref_fasta_filename%.gz}"
wget -q -O - "${ref_fasta}"| zcat > "$ref_fasta_filename"
```

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
...
+
+    ref_fasta_filename=$(echo "$ref_fasta_tar_filename" | rev | cut -c4- | rev)
+
+    wget ${gtf_file}
+    gtf_tar_filename=$(echo ${gtf_file} | sed 's:.*/::')
+    gunzip $gtf_tar_filename
+
+    gtf_filename=$(echo "$gtf_tar_filename" | rev | cut -c4- | rev)
```
mckinsel: I would use the pattern above for the fasta for the gtf as well. If this is notably slow, you can probably perform the two commands in parallel.

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
...
+
+workflow BuildCellRangerRef {
+  String fasta
+  String gtf
+  String ref_name
+
+  call BuildCellRangerReference {
+    input:
+      ref_fasta = fasta,
+      gtf_file = gtf,
```
mckinsel: Any reason not to have these names be consistent across workflow and task? IMO makes it easier to follow.

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
mckinsel: If you do end up gzipping, consider `tar cvf - GRCh38/ | pigz > "${ref_name}.tar.gz"`

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
...
+
+    ref_fasta_filename=$(echo "$ref_fasta_tar_filename" | rev | cut -c4- | rev)
+
+    wget ${gtf_file}
+    gtf_tar_filename=$(echo ${gtf_file} | sed 's:.*/::')
+    gunzip $gtf_tar_filename
+
+    gtf_filename=$(echo "$gtf_tar_filename" | rev | cut -c4- | rev)
+
+    cellranger mkgtf $gtf_filename "filtered.$gtf_filename" \
```
mckinsel: Can you include a link to https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/advanced/references somewhere? Might be helpful to let people know where this is coming from.

mckinsel: Also, can you double-quote `"$gtf_filename"`?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,80 @@
...
+  String ref_name
+
+  String docker = "quay.io/humancellatlas/secondary-analysis-cellranger"
+  String memory = "416 GB"
+  Int boot_disk_size_gb = 12
+  String disk_space = "250"
+  Int cpu = 64
+
+  command {
+    set -e
```
mckinsel: Probably also want an `-o pipefail` given the use of pipelines.

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
jishuxu: The index files are binary files, I am wondering how much you can reduce in size if you compress.

rexwangcc:
@dshiga @mckinsel I made the changes based on your comments, would you mind taking another look at the PR? Thanks!

mckinsel:
Thanks @rexwangcc, looks good.

kbergin:
LGTM!

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,102 @@
...
+                     --fasta="$ref_fasta_filename" \
+                     --genes="filtered.$gtf_filename" \
+                     --nthreads=$(getconf _NPROCESSORS_ONLN)
+
+    tar cvf "~{ref_name}.tar" GRCh38/
+
+  >>>
+
+  runtime {
+    bootDiskSizeGb: boot_disk_size_gb
```
dshiga: Why are we specifying this here? We don't do it in other pipelines, do we?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,102 @@
+version 1.0
+
+
+workflow BuildCellRangerRef {
```
dshiga: Could you add a one line meta description of what this workflow is for?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
rexwangcc: We haven't done this in other pipelines, except the cellranger one: https://github.com/HumanCellAtlas/skylab/blob/ffd17882c831da377dce3ab471c7a44683dd00f6/pipelines/cellranger/cellranger.wdl#L16 Don't know if it's a required, but just copied over

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
rexwangcc: Oops, I thought I added that, will do!

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -0,0 +1,102 @@
...
+    docker: "(optional) the docker image containing the runtime environment for this task"
+    boot_disk_size_gb: "(optional) the amount of space (GB) of boot disk to provision for this task"
+    cpu: "(optional) the number of cpus to provision for this task"
+    memory: "(optional) the amount of memory (MB) to provision for this task"
+  }
+
+  command <<<
+
+    set -euo pipefail
+
```
dshiga: Could you add some comments to summarize the code below? It's a little dense. Looks like it's copying down the file and gunzipping it?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
dshiga: @samanehsan can you comment on why the cellranger pipeline specifies this?

samanehsan: @dshiga this was a part of the cellranger.wdl that was in firecloud, and I kept the same parameters

rexwangcc: @dshiga I found it require a lot of space on our boot disk(even you have a large mounted disk) when you pull the cellranger docker image from quay.io The docker daemon will complain and exit unexpectedly. So I prefer we keep this parameter here to provide some flexibility to override it from input json files.

dshiga: Great, thanks for looking into it @rexwangcc. Would you mind adding a code comment explaining that this parameter is here because the cellranger image is so large?

rexwangcc: Will do!

jishuxu: Just for the record, I think we need to increase `bootDiskSizeGb` if the  programs(java application, for example) we run require a lot of TMP space.

rexwangcc: @jishuxu That makes sense! Another reason to keep it configurable here!

rexwangcc:
I'm going to merge this PR!

PR #132, opened 2018-10-03
Jx cellranger build ref gencode

jishuxu:
In this PR: 
Add a wdl workflow to generate cellranger reference using gencode annotation


mckinsel:
Hey what's the intersection between the count matrix columns produced by this and the ones in SS2?

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundleGencode.wdl
```diff
@@ -0,0 +1,123 @@
+version 1.0
+
+
+workflow BuildCellRangerRef {
```
kbergin: Could this workflow name and the task name below be distinguished from the ensembl version of this workflow? The wdl filename here makes it clear that it's Gencode but the workflows and tasks are then named the same. Maybe `BuildCellRangerGencodeRef`

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundleGencode.wdl
```diff
@@ -0,0 +1,123 @@
...
+  meta {
+    description: "Accessory workflow for builing the reference for CellRanger pipeline."
```
kbergin: Similarly, could this description mention that this is for Gencode? thanks!

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundleGencode.wdl
```diff
@@ -0,0 +1,123 @@
...
+      disk_space = disk_space,
+      cpu = cpu
+  }
+
+  output {
+    File cellranger_ref = BuildCellRangerReference.cellRangerRef
+  }
+}
+
+task BuildCellRangerReference {
```
kbergin: Same comment as above. Thanks!

jishuxu:
@mckinsel 
> Hey what's the intersection between the count matrix columns produced by this and the ones in SS2?
Did you mean the gene_id or transcript_id columns in RSEM outputs of SS2?

mckinsel:
@jishuxu, I think I mean the gene id. (I think transcript doesn't make sense for 3' method right?)

jishuxu:
> @jishuxu, I think I mean the gene id. (I think transcript doesn't make sense for 3' method right?)

Just keep in mind that RSEM's output only have gene_id or tracking_id, not gene name.

samanehsan:
@jishuxu it looks like the BuildCellRangerReferenceBundleEnsembl.wdl and BuildCellRangerReferenceBundleGencode.wdl are the same. Can we just have one generic wdl in that case and keep the two different input JSON files?

jishuxu:
> @jishuxu it looks like the `BuildCellRangerReferenceBundleEnsembl.wdl` and `BuildCellRangerReferenceBundleGencode.wdl` are the same. Can we just have one generic wdl in that case and keep the two different input JSON files?

I think the main workflow is the same but the task and inputs are the different. One is use `gene_type` and another is `gene_biotype` for attribute name. I talked with @rexwangcc before I implemented. my original thought was to use a input parameter to trigger the different task in the same workflow. The current way just tried to reduce the confusion. I am open to either approach,  let's pick up the best way then.

samanehsan:
Oh I see! Could you make the attribute a parameter if that is the only difference between the two tasks?

jishuxu:
> Oh I see! Could you make the attribute a parameter if that is the only difference between the two tasks?

Sure, I will try!

jishuxu:
Just updated and re-factor-ed wdls. Here is list of changes.
1) Now, to build `Ensembl` and `Gencode` references by one workflow `BuildCellRangerRef`
2) add a input variable `ref_source` to differentiate different annotation, `Ensembl` and `Gencode`
3) Also updated runtime. Now it only requires 8-core and 30G ram standard machine. It does takes a little bit longer compared to 64 cores. 
4) two jsons files as inputs, `BuildCellRangerReferenceBundleEnsembl.json` and `BuildCellRangerReferenceBundleGencode.json`

library/accessory_workflows/build_cellranger_reference/BuildCellRangerReferenceBundle.wdl
```diff
@@ -79,8 +81,8 @@ task BuildCellRangerReference {
 
     # The reference building process follows the instructions from
     # https://support.10xgenomics.com/single-cell-gene-expression/software/pipelines/latest/advanced/references
-
-    cellranger mkgtf "$gtf_filename" "filtered.$gtf_filename" \
+    if [[ "~{ref_source}"=="Ensembl" ]]; then
```
samanehsan: If the parameters to the cellranger mkgtf call will only vary by the attribute type, you could do something like this to consolidate it a bit more:

```
if [[ "~{ref_source}"=="Ensembl" ]]; then
     attribute_type="gene_biotype"
elif [[ "~{ref_source}"=="Gencode" ]]; then
     attribute_type="gene_type"
fi
cellranger mkgtf "$gtf_filename" "filtered.$gtf_filename" \
              --attribute=${attribute_type}:protein_coding \
              --attribute=${attribute_type}:lincRNA \
              ....
```

samanehsan: Oh, okay! That makes sense to leave it as-is then

samanehsan: @jishuxu can you please make the indentation consistent here?

jishuxu: is it better now?

samanehsan: Yes, thank you!

kbergin:
Btw I love seeing the refactor of the code here, it is much cleaner now :) Thanks!!

PR #134, opened 2018-10-06
Fix disk space issue for hisat2

dshiga:
Changed disk space formula to scale better with larger outputs, especially those with many multi-mapped reads. Also modified hisat2 commands to convert output to bam on the fly so no sam file needs to be written (this reduces disk space required for outputs by factor of up to ~10).

See https://github.com/HumanCellAtlas/secondary-analysis/issues/361

- [ ] Added or updated tests
- [x] Updated documentation

### DIFF
--- a/library/tasks/HISAT2.wdl
+++ b/library/tasks/HISAT2.wdl
@@ -10,8 +10,9 @@ task HISAT2PairedEnd {
   String docker = "quay.io/humancellatlas/secondary-analysis-hisat2:v0.2.2-2-2.1.0"
   Int machine_mem_mb = 15000
   Int cpu = 4
-  # use provided disk number or dynamically size on our own, 10 is our zipped fastq -> bam conversion with 50GB of additional disk
-  Int disk = ceil((size(fastq1, "GB") + size(fastq2, "GB") * 10) + size(hisat2_ref, "GB") + 50)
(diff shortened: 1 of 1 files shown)


library/tasks/HISAT2.wdl
```diff
@@ -71,8 +72,8 @@ task HISAT2PairedEnd {
       --seed 12345 \
       -k 10 \
       --secondary \
-      -p ${cpu} -S ${output_basename}.sam
-    samtools sort -@ ${cpu} -O bam -o "${output_basename}.bam" "${output_basename}.sam"
+      -p ${cpu} -S >(samtools view -1 -h -o ${output_basename}_unsorted.bam)
```
mckinsel: Instead of hard coding the parallelism, maybe use something like `$(nproc)`. I don't think executors are necessarily required to honor the `runtime.cpu` value.

rexwangcc: Just a side note that `nproc` might not exist in some docker images.

dshiga: I'll create a ticket for this to be addressed in a future PR.

dshiga: https://app.zenhub.com/workspace/o/humancellatlas/secondary-analysis/issues/363

PR #140, opened 2018-10-16
Set correct tag for image in zarr task

mckinsel:
- [ ] Added or updated tests
- [ ] Updated documentation


samanehsan:
Thanks @mckinsel! :octocat:

PR #144, opened 2018-10-30
Add max_retries param to cellranger

samanehsan:
Add max_retries parameter to the CellRanger task so that it retries if the task fails. This will add robustness against intermittent google cloud failures, such as the VM quitting unexpectedly.


samanehsan:
The portability test does not run the Cellranger WDL at the moment, so I think the failure is related to the minor service outage that just occurred

PR #160, opened 2019-02-01
Add a single-end SmartSeq2 pipeline

mckinsel:
This copies the logic from the paired-end pipeline but adjusts the
command invocations to handle single-end reads.

Note that this also requires a small change to sctools.


mckinsel:
Sure no problem, I will fork and move this there. But your plan is just indefinite delay, which isn't in the charter either!

PR #165, opened 2019-02-22
Kmk fixed chunk size zarr

kishorikonwar:
### Purpose
The zarr conversion of Optimus output of the was running out of memory in the expression matrix conversion (to zarr) step. This is because the chunk size of the matrix was not fixed and therefore depending on the samples number of barcodes or genes it would vary, sometimes the size of each chunk is the max RAM it requires.

## Changes
Now, it is changed to a size of 10000 x 10000, and the logic of the code is changed to convert the expression matrix to zarr chunk by chunk (or tile by tile) of size 10000x 10000

### Review Instructions
Review the logic of the code where it has changed.

### Follow-up Discussions
* No follow-up discussions.


docker/python3-scientific/create_zarr_optimus.py
```diff
@@ -282,18 +283,25 @@ def add_expression_counts(data_group, args):
...
     # load a chunks from the expression count matrix and update the corresponding chunk in expression matrix in zar
     for i in range(0, csr_exp_counts.shape[0], CHUNK_ROW_SIZE):
-        # check if it is possible to make a full chunk of data
-        if i + CHUNK_ROW_SIZE <= csr_exp_counts.shape[0]:
-            exp_counts_group[i:i + CHUNK_ROW_SIZE, ] = csr_exp_counts[i:i + CHUNK_ROW_SIZE, ].toarray()
-        else:
-            # not enough data remaining, so make the final set of data less than a chunk
-            j = csr_exp_counts.shape[0] - i
-            exp_counts_group[i:i + j, ] = csr_exp_counts[i:i + j, ].toarray()
+      for j in range(0, csr_exp_counts.shape[1], CHUNK_COL_SIZE):
```
rexwangcc: Please fix the indentations:
```python
for j in range(0, csr_exp_counts.shape[1], CHUNK_COL_SIZE):
    # check if it is possible to make a full row chunk of data, otherwise adjust to the correct size
    p= CHUNK_ROW_SIZE
    if i + CHUNK_ROW_SIZE > csr_exp_counts.shape[0]:
        p = csr_exp_counts.shape[0] - i

    # check if it is possible to make a full row chunk of data, otherwise adjust to the correct size
    q = CHUNK_COL_SIZE
    if j + CHUNK_COL_SIZE > csr_exp_counts.shape[1]:
        q = csr_exp_counts.shape[1] - j

    #insert the chunk
    exp_counts_group[i:i + p, j:j + q] = csr_exp_counts[i:i + p, j: j + q].toarray()
```

docs/matrix_format_spec.md
```diff
@@ -23,7 +23,7 @@ UUID generated by the DCP or a user-submitted identifier.
   metadata arrays defined below.
   - shape: `(n_cells, n_genes)`
   - compressor: `Blosc(cname='lz4', clevel=5, shuffle=SHUFFLE, blocksize=0)`
-  - chunk shape: `(10000, n_genes)`
```
rexwangcc: should we also change line 43, since it's also `chunk shape: (n_genes,)`?

kishorikonwar:
retest this please

kishorikonwar:
retest this please

PR #166, opened 2019-02-22
Add scripts for generating test dataset by genomic location [WIP]

barkasn:
### Purpose
This PR adds scripts and wdl files as well as the required and docker containers for generating test datasets for optimus, such that all reads in the input fastqs will align only to given genomic locations.


docker/subset-fastq-dataset/prepSubset.sh
```diff
@@ -0,0 +1,118 @@
...
+###########################
+## Internal params
+keepreadsgz=keepreads.txt.gz
+samtoolsSortCores=4
+samtoolsMemPerCore=5G
+
+## Get the aligned files
+echo Downloading alignments...
+mkdir starAlignOut
+gsutil cp -r $starAlignOutBucket starAlignOut/
```
pshapiro4broad: I would use `gsutil -m` here and on line 90 below so the multiple file copy operation is multithreaded.

docker/subset-fastq-dataset/prepSubset.sh
```diff
@@ -0,0 +1,118 @@
...
+samtools cat `find ./starAlignOut/ -name '*.bam' -printf '%p '` > concat.bam
+
+## Sort and index the concatenated file
+echo Sorting merged alignments...
+samtools sort -@ ${samtoolsSortCores} -m ${samtoolsMemPerCore} concat.bam > concat.sorted.bam
+samtools index concat.sorted.bam
+
+## Extract read names
+echo Extracting read names for $samtoolsRegion
+samtools view concat.sorted.bam $samtoolsRegionn | cut -f 1 | pigz > $keepreadsgz
```
infispiel: This seems to be a typo? Regionn instead of Region?

docker/subset-fastq-dataset/build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+tag=$1
+
+if [ -z $tag ]; then
```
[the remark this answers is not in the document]
barkasn: The tag here is the docker image tag. Not sure why it would be in hts specs. This is a convention we started with Rex because previously the image building was done manually that was very error prone and it was also rather difficult to find the correct tag for each image.

infispiel:
Note that I put absolutely 0 effort towards tweaking the runtime parameters on a per-task basis, so once I've confirmed that the pipeline runs correctly, I'd really appreciate Nick or Kishori tweaking those to be as much as the step needs.

infispiel:
Currently running commit 8cb7577 with this ID:
0662604a-c063-4c5e-b8b6-e72a7e69e813

On success the wdl should be good to go; the following commit should have little to no effect on success/failure.

docker/subset-fastq-dataset/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" |  tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg |  apt-key add -
+RUN apt update
+RUN apt install -y google-cloud-sdk
+
+## Install latest samtools from source
+WORKDIR /tools
+RUN wget https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2 && tar xvjf samtools-1.9.tar.bz2
+WORKDIR /tools/samtools-1.9
+RUN ./configure && make -j 4 && cp samtools ..
```
tlangs: This whole run should be 
```bash
RUN mkdir /tools && \
  cd /tools && \
  wget https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2 && \
  tar xvjf samtools-1.9.tar.bz2 && \
  cd samtools-1.9 && \
  ./configure && \
  make -j 4 && \
  cp samtools ..
```
This way, only one interim image is created for the install of samtools.

docker/subset-fastq-dataset/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+
+## Install pigz
+RUN apt update && apt install -y pigz
+
+## Copy Scripts
+COPY filterFastqByReadName.py /tools/
+COPY prepSubset.sh /tools/
```
tlangs: `WORKDIR` will be the directory that the image boots into, so it should be set to where the user can immediately use the tools provided. In this case, it looks like `/tools`.

docker/subset-fastq-dataset/filterFastqByReadName.py
```diff
@@ -0,0 +1,33 @@
...
+@click.option('--keep-reads-gz',help='gz compressed file with read names to keep')
+@click.option('--verbose',help='verbose',default=False,is_flag=True,flag_value=True)
+def filterbyreadname(in_fastq_gz, out_fastq_gz, keep_reads_gz, verbose):
+    import gzip
+    import sys
+    from Bio import SeqIO
+    ## Put reads that are in the keep list in the output file
+    keep_read_set = set(line.decode('ascii').rstrip() for line in gzip.open(keep_reads_gz,'rb'))
+    if (verbose):
+        print('Done loading keep read list', file=sys.stderr);
```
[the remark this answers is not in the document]
tlangs: Outputting to `stderr` will not cause `set -e` to fail. Only a return code not `0` will cause `set -e` to trigger.

library/accessory_workflows/subset_fastq_dataset/subset_fastq_dataset.wdl
```diff
@@ -0,0 +1,149 @@
...
+
+task ExtractReadNames {
+    File sorted_bam
+    File sorted_bai
+    String subset_region
+    Int samtools_sort_cores = 4
+    String samtools_mem_per_core = "5G"
+
+    command {
+        samtools view -@ ${samtools_sort_cores} -m ${samtools_mem_per_core} ${sorted_bam} ${subset_region} | cut -f 1 | pigz > kept_reads.gz
```
tlangs: Why are we providing a gzipped file as output instead of an `Array[String]`. Using an `Array[String]` would mean that the output could be provided to the next task directly and there would be no file to localize.

library/accessory_workflows/subset_fastq_dataset/subset_fastq_dataset.wdl
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    call TarFilteredFastq {
+        input:
+        filtered_fastqs = FilterFastq.output_filtered_fastq_gz
+    }
+
+    output {
+        File output_tar = TarFilteredFastq.output_tarred_fastqs
```
tlangs: How come the output of this workflow is a tarred file? An `Array[File]` would mean that we no longer have to include `pigz` in the docker image and get rid of the untarring steps in tasks. It also means that the output files can be looked at individually instead of having to download an entire dataset. Unless this is a requirement for the WDL, I would suggest that we don't compress files in the WDL to keep things as easy to comprehend as possible.

docker/subset-fastq-dataset/filterFastqByReadName.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+
+import click
+
+@click.command()
+@click.option('--in-fastq-gz',help='gz compressed input fastq file')
+@click.option('--out-fastq-gz',help='gz compressed output fastq file')
+@click.option('--keep-reads-gz',help='gz compressed file with read names to keep')
```
tlangs: Is there a reason everything is gzipped?

docker/subset-fastq-dataset/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+    numpy==1.16.2 \
+    pysam==0.15.2 \
+    biopython==1.73
+
+## Make directory for tools and put them in the path
+RUN mkdir /tools
+ENV PATH=/tools/:${PATH}
+
+## Note that this string is specific this linux base
+ENV CLOUD_SDK_REPO="cloud-sdk-jessie"
```
tlangs: Does this need to be an environment variable or can it just be inlined in the below `RUN` usage?

library/accessory_workflows/subset_fastq_dataset/subset_fastq_dataset.wdl
```diff
@@ -0,0 +1,149 @@
...
+    command {
+        outfastqgz=${output_file_name}
+        filterFastqByReadName.py \
+            --in-fastq-gz ${fastq} \
+            --out-fastq-gz $outfastqgz \
+            --keep-reads-gz ${keep_read_names_gz} \
+            --verbose
+    }
+
+     runtime {
```
tlangs: weird indentation for the runtime block. Looks like there's an extra space.

library/accessory_workflows/subset_fastq_dataset/subset_fastq_dataset.wdl
```diff
@@ -0,0 +1,149 @@
...
+}
+
+task TarFilteredFastq {
+    Array[File] filtered_fastqs
+
+    command {
+        tar -cvzf output.tar.gz ${sep=' ' filtered_fastqs}
+    }
+
+     runtime {
```
tlangs: Same extra space in indentation.

docker/subset-fastq-dataset/Dockerfile
barkasn: It's easier to change in the future if it is standalone, if the base ubuntu image is ever changed this will require updating as well.

docker/subset-fastq-dataset/filterFastqByReadName.py
barkasn: I've always found that compressing fastqs almost always leads to performance increase because reading/writing such files when doing light processing is IO bound. Also helps prevent filling up the HDD.

library/accessory_workflows/subset_fastq_dataset/subset_fastq_dataset.wdl
barkasn: Can be improved, but haven't worked with Arrays in WDL before so not confident to fix this.

docker/subset-fastq-dataset/prepSubset.sh
barkasn: done

docker/subset-fastq-dataset/prepSubset.sh
barkasn: fixed, probably the reason my runs were failing because without this the whole file is output

docker/subset-fastq-dataset/Dockerfile
barkasn: done

library/accessory_workflows/subset_fastq_dataset/subset_fastq_dataset.wdl
barkasn: Can be improved, but haven't worked with Arrays in WDL before so not confident to fix this.

docker/subset-fastq-dataset/Dockerfile
barkasn: What is it by default?

docker/subset-fastq-dataset/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN apt install -y google-cloud-sdk
+
+## Install latest samtools from source
+WORKDIR /tools
+RUN wget https://github.com/samtools/samtools/releases/download/1.9/samtools-1.9.tar.bz2 && tar xvjf samtools-1.9.tar.bz2
+WORKDIR /tools/samtools-1.9
+RUN ./configure && make -j 4 && cp samtools ..
+
+## Install pigz
+RUN apt update && apt install -y pigz
```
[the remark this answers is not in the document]
barkasn: done

docker/subset-fastq-dataset/Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+RUN mkdir /tools
+ENV PATH=/tools/:${PATH}
+
+## Note that this string is specific this linux base
+ENV CLOUD_SDK_REPO="cloud-sdk-jessie"
+
+RUN echo "deb http://packages.cloud.google.com/apt $CLOUD_SDK_REPO main" |  tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg |  apt-key add -
+RUN apt update
+RUN apt install -y google-cloud-sdk
```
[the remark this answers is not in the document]
barkasn: done

docker/subset-fastq-dataset/filterFastqByReadName.py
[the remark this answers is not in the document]
barkasn: I think Trevyn is right, -e fails on exit code other than 0. Verbosity doesn't matter. Here is a reference link: https://www.gnu.org/software/bash//manual/html_node/The-Set-Builtin.html

docker/subset-fastq-dataset/run.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+./prepSubset.sh \
+    -a gs://broad-dsde-mint-dev-cromwell-execution/cromwell-executions/Optimus/4a43c049-39bc-45be-a7a4-fac7e4dc6a41/call-StarAlign \
```
als364: do your cromwell execution buckets have a TTL on them?

barkasn: But, we actually don't need this file

als364: Do you mean the file in the broad-dsde-mint-dev-cromwell-execution bucket or this run.sh file?

PR #167, opened 2019-02-27
Phs add preemptible where sensible

infispiel:
### Purpose
This PR contains changes that sets Optimus to default run all tasks as preemptible, with an option to run all tasks as non-preemptible.

---
### Changes

- All tasks called directly by `Optimus.wdl` have had their default preemptible value set to 3. (This is technically extraneous, since `Optimus.wdl` now always passes a `preemptible` value to every task called.)
- The `preemptible` variable has been added to `ZarrUtils.wdl > SmartSeq2ZarrConversion` & `ZarrUtils.wdl > OptimusZarrConversion`. The `preemptible` variable has previously been described in the `parameter_meta` section of both of these tasks but was not actually implemented.
- `Optimus.wdl` now also takes an optional parameter `preemptible` that sets the preemptibility of all tasks called by the Optimus pipeline. The default value is 3. Set to 0 to set as non-preemptible.
- A brief explanation of the new `preemptible` option has been added to the README as I, personally, thought it had non-obvious consequences to a user who understands how to run WDL pipelines.
---
### Review Instructions

- Run the pipeline without setting a `preemptible` value.
- Do a metadata query in Swagger with `preemptible` in the includeKey field.
- Ensure that all tasks were attempted with `"preemptible": true`. If `"preemptible": false` appears, I believe it should appear with the value `"attempt": 4` underneath it to prove it was attempted on preemptible machines at least 3 times. Alternatively, ensure that that task with the same shard ID has appeared as `"preemptible": true`.

---
### Follow-up Discussions

- I set the `preemptible` default value to be 3 on all tasks called by Optimus. There were a few task files that were not modified. It may be worth changing these to `preemptible = 3` as well, but I was focusing on making the Optimus pipeline preemptible.
- Alternatively, it may be worth re-setting the `preemptible = 0`, as Optimus now always passes a `preemptible` value to these tasks.
- I did not modify the default preemptible values on Picard despite the fact it is called by Optimus, as these were already set to 5. These could be set to 3 for consistency, or left alone, as again--Optimus passes it a `preemptible` value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,12 @@ For now, use `git clone git@github.com:HumanCellAtlas/skylab.git` and run the pi
 1. WDL and Cromwell [Documentation](https://software.broadinstitute.org/wdl/)
 2. Running WDLs in [Cromwell](https://software.broadinstitute.org/wdl/documentation/execution.php)
 
+### preemptible
+
+Tasks on Cromwell may be run on what are known as "preemptible" machines to reduce costs by a significant amount. The catch with preemptible machines is that they may be "preempted" at any given moment--as in, google may shut down the task to re-use the resources.
+
+Many tasks are set to automatically be `preemptible = 3`, aka they will be run on preemptible instances for up to 3 instances of preemption, after which it will be run on a non-preemptible machine. The Optimus pipeline in particular also has a `preemptible` input that overrides the default preemptible parameter on all tasks run through optimus. This option may be set to 0 to run the entire workflow without using preemptible machines.
+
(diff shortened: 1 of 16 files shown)


infispiel:
I found a few places that need more preemptible passes from Optimus to tasks--please hold on reviewing. I will have a commit hopefully within a few minutes.

infispiel:
Workflow ID: 972f86e5-225a-408e-bd15-bc04bf1ef871
All steps were successful.

PR #172, opened 2019-03-13
updating the skylab pull approve yaml to v3

rhiananthony:
### Purpose
The purpose of this PR is to update PullApprove to use v3

---
### PR Checklist
_Please ensure the following when opening a PR:_

- [ ] This PR added or updated tests.
- [ ] This PR updated docstrings or documentation.
- [ ] Let Jodi (**jodihir**) know when there is user-facing documentation updates needed!
- [ ] This PR applied the Mint style guide for WDL.
- [ ] This PR has no WDL linting errors reported by [miniwdl](https://github.com/chanzuckerberg/miniwdl)
- [ ] This PR considered generalizability beyond our own use case.

---
### Follow-up Discussions

- No follow-up discussions.


.pullapprove.yml
```diff
@@ -1,45 +1,44 @@
...
+
+    reviewers:
+      users:
+      - rhiananthony
+      teams:
+      - secops
+
+  software-review:
+    reviews:
+      required: 2
```
samanehsan: Why are two software reviews required? I thought it was one software and one computational review

rhiananthony: This requires 2 software reviews unless the author is part of that group, then it only requires one. For instance, if anyone on hornet creates a PR then it only requires one more SE and one CB approval. If someone else outside of this group decided to create a PR - Sam for instance, then it would require review by 2 SEs and one CB. Does that make sense? I can make it more simple, but was experimenting :)

samanehsan: Oh I see! I vote keeping it one SE and one CB approval even for external changes to skylab

rhiananthony: Done :)

PR #173, opened 2019-03-19
Disable preemptible for some time-consuming tasks.

rexwangcc:
### Purpose
_Please explain the purpose of this PR and include links to any GitHub issues that it fixes:_

- https://broadinstitute.atlassian.net/jira/software/projects/GL/boards/528?selectedIssue=GL-166
- Just looking at this run: https://job-manager.caas-prod.broadinstitute.org/jobs/b3f61b06-635c-466e-997c-2a77a313c6d1?q=caas-collection-name%3Dlira-test%26bundle-uuid%3D00ba4f21-7a70-4517-8bdd-a9ac9aa4c444 we can tell that it's not practical to use the same `preemptible` settings for all of the tasks in Optimus. The above run was only on a very small dataset, if we run Optimus on larger datasets, things will get worse. We need finer control over the settings.

---
### Changes
_Please list out what major changes were made in this PR to address the issue:_

- Add another `no_preemptible` switch to the Optimus pipeline, so we can disable `preemptible` on specific steps. Downside: not atomic enough. Especially when you run the Optimus within an Adapter workflow, there's no way you can change the pre-defined settings except modifying the pipeline itself.

---
### Review Instructions
_Please provide instructions about how should a reviewer test/verify the changes in this PR:_

- This PR is mutually exclusive to the https://github.com/HumanCellAtlas/skylab/pull/174, only one of them can be merged🥇 ! 

---
### PR Checklist
_Please ensure the following when opening a PR:_

- [ ] This PR added or updated tests.
- [ ] This PR updated docstrings or documentation.
- [ ] Let Jodi (**jodihir**) know when there is user-facing documentation updates needed!
- [ ] This PR applied the Mint style guide for WDL.
- [ ] This PR has no WDL linting errors reported by [miniwdl](https://github.com/chanzuckerberg/miniwdl)
- [ ] This PR considered generalizability beyond our own use case.

---
### Follow-up Discussions

- No follow-up discussions.

### DIFF
--- a/pipelines/optimus/Optimus.wdl
+++ b/pipelines/optimus/Optimus.wdl
@@ -59,8 +59,7 @@ workflow Optimus {
       input:
         fastq_file = r2_fastq[index],
         sample_id = sample_id,
-        fastq_suffix = fastq_suffix,
-        preemptible = preemptible
+        fastq_suffix = fastq_suffix
     }
 
     # if the index is passed, attach it to the bam file
@@ -71,8 +70,7 @@ workflow Optimus {
           r1_fastq = r1_fastq[index],
           i1_fastq = non_optional_i1_fastq[index],
           r2_unmapped_bam = FastqToUBam.bam_output,
-          whitelist = whitelist,
-          preemptible = no_preemptible
+          whitelist = whitelist
       }
     }
 
@@ -82,8 +80,7 @@ workflow Optimus {
         input:
           r1_fastq = r1_fastq[index],
           r2_unmapped_bam = FastqToUBam.bam_output,
-          whitelist = whitelist,
-          preemptible = no_preemptible
+          whitelist = whitelist
       }
     }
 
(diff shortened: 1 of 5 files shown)


infispiel:
It seems a bit odd to set a `no_preemptible` variable that is passed to these tasks--unless you're planning to change `no_preemptible`, in which case the name choice is inaccurate. I have 2 suggestions, if you'd like to consider them:
1. Don't pass any preemptible value within Optimus, and set the default `preemptible` to 0 in the tasks WDLs themselves. This will have them default to 0 unless specifically passed as an input.
2. Pass a `Bool force_non_preemptible` so if it is set to false, the preemptible value is used, and if set to true, preemptible for those tasks is set to 0.

pipelines/optimus/Optimus.wdl
```diff
@@ -75,7 +72,7 @@ workflow Optimus {
           i1_fastq = non_optional_i1_fastq[index],
           r2_unmapped_bam = FastqToUBam.bam_output,
           whitelist = whitelist,
-          preemptible = preemptible
+          preemptible = no_preemptible
```
infispiel: I don't see a no_preemptible variable set anywhere?

rexwangcc: Good catch, Cromwell just reported it too...  😢 will fix right away.

rexwangcc: Also I will remove all of the `preemptible = preemptible`

pipelines/optimus/Optimus.wdl
```diff
@@ -41,8 +41,6 @@ workflow Optimus {
   # whether to convert the outputs to Zarr format, by default it's set to true
   Boolean output_zarr = true
 
-  Int preemptible = 3
```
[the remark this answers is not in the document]
rexwangcc: Fixing

rexwangcc:
The failed test has nothing to do with this PR, I'm going to merge it!

PR #178, opened 2019-03-29
Update LICENSE

kbergin:
2017->2019

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2017 Human Cell Atlas Authors, https://humancellatlas.org
+Copyright (c) 2019 Human Cell Atlas Authors, https://humancellatlas.org
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright (c) 2017 Human Cell Atlas Authors, https://humancellatlas.org
+Copyright (c) 2019 Human Cell Atlas Authors, https://humancellatlas.org
```
mckinsel: Think you want a range or list here: `2017, 2018, 2019`

kbergin: Huh that does seem possible. I just assumed it should be current. Perhaps @TimothyTickle knows better than I do.

kbergin: @TimothyTickle do you happen to know the answer here?

rexwangcc: I don't know the answer but did some googling: https://softwareengineering.stackexchange.com/a/210491

ambrosejcarr:
```suggestion
Copyright (c) 2017-2019 Human Cell Atlas Authors, https://humancellatlas.org
```
Based on Rex and Marcus's comments, the suggested change appears to be a correct approach.

nikellepetrillo:
Closing this and will update in the new WARP repo: https://github.com/broadinstitute/warp

PR #184, opened 2019-04-14
Upload example contribution PR using circle

ambrosejcarr:
### Purpose

Contribution example to skylab that adds a minimal workflow and check task that adds a test for itself using @mckinsel 's circle-based s3/local cromwell backend. 

### Review Instructions
Look at the CI `build` test results. 

### PR Checklist
_Please ensure the following when opening a PR:_

- [X] This PR added or updated tests.
- [ ] This PR updated docstrings or documentation.
- [ ] Let Jodi (**jodihir**) know when there is user-facing documentation updates needed!
- [X] This PR applied the Mint style guide for WDL.
- [X] This PR has no WDL linting errors reported by [miniwdl](https://github.com/chanzuckerberg/miniwdl)
- [X] This PR considered generalizability beyond our own use case.

---
### Follow-up Discussions
_Please append follow-up discussions and issues during the review process below:_

- No follow-up discussions.

PR #195, opened 2019-05-10
Add optimus test to circle ci

jessicaway:
### Purpose

Replacing the Jenkins optimus test with a circle ci optimus test

### Changes

added optimus test to circle ci

changed the jenkin's tests that were run on mint-dev cromwell to run on caas-prod cromwell

### Review Instructions

- No instructions.

### DIFF
--- a/test/test_cromwell_workflow.sh
+++ b/test/test_cromwell_workflow.sh
@@ -2,26 +2,27 @@
 
 set -e
 
-CROMWELL_USER=$1
-CROMWELL_PASSWORD=$2
-CROMWELL_URL=$3
-INPUTS_JSON=$4
-WDL_FILE=$5
-OPTIONS_FILE=$6
-DEPENDENCIES_JSON=$7
+CROMWELL_KEY_FILE=$1
+CROMWELL_URL=$2
+INPUTS_JSON=$3
+WDL_FILE=$4
+OPTIONS_FILE=$5
+DEPENDENCIES_JSON=$6
+COLLECTION=$7
 # Read list of dependency files into an array to pass to cromwell-tools
 DEPENDENCIES_LIST=$(cat ${DEPENDENCIES_JSON} | python3 -c "import json,sys;obj=json.load(sys.stdin);print(' '.join(obj.values()));")
 echo ${DEPENDENCIES_LIST[@]}
 
 echo "Starting workflow."
+
 WORKFLOW_HASH=$(cromwell-tools submit \
-  --username "${CROMWELL_USER}" \
-  --password "${CROMWELL_PASSWORD}" \
+  --service-account-key "${CROMWELL_KEY_FILE}" \
   --url "${CROMWELL_URL}" \
   --inputs_files "${INPUTS_JSON}" \
   --wdl-file "${WDL_FILE}" \
   --options-file "${OPTIONS_FILE}" \
   --deps-file ${DEPENDENCIES_LIST[@]} \
+  --collectiuon-name ${COLLECTION} \
 )
 
 # Get workflow id from cromwell-tools response: {"id": "XXXXX", "status": "Submitted"}
@@ -35,16 +36,14 @@ sleep 10
 
 echo "Waiting for workflow ${WORKFLOW_ID} to complete..."
 cromwell-tools wait "${WORKFLOW_ID}" \
-  --username "${CROMWELL_USER}" \
-  --password "${CROMWELL_PASSWORD}" \
+  --service-account-key "${CROMWELL_KEY_FILE}" \
   --url "${CROMWELL_URL}" \
   --timeout-minutes 60 \
   --poll-interval-seconds 30
 
 echo "Checking workflow completion status."
 STATUS_RESULT=$(cromwell-tools status \
-  --username "${CROMWELL_USER}" \
-  --password "${CROMWELL_PASSWORD}" \
+  --service-account-key "${CROMWELL_KEY_FILE}" \
   --url "${CROMWELL_URL}" \
   --uuid "${WORKFLOW_ID}" \
 )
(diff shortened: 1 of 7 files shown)


test/test_cromwell_workflow.sh
```diff
@@ -2,26 +2,27 @@
...
 WORKFLOW_HASH=$(cromwell-tools submit \
-  --username "${CROMWELL_USER}" \
-  --password "${CROMWELL_PASSWORD}" \
+  --service-account-key "${CROMWELL_KEY_FILE}" \
   --url "${CROMWELL_URL}" \
   --inputs_files "${INPUTS_JSON}" \
   --wdl-file "${WDL_FILE}" \
   --options-file "${OPTIONS_FILE}" \
   --deps-file ${DEPENDENCIES_LIST[@]} \
+  --collectiuon-name ${COLLECTION} \
```
ambrosejcarr: Can you investigate if this is being called? There's a spelling error here that makes me question if this does anything.

test/trigger_test.sh
```diff
@@ -4,34 +4,35 @@ set -e
...
-INPUTS_JSON=$2
-WDL_FILE=$3
-DEPENDENCIES_JSON=$4
+PIPELINE_FOLDER_NAME=$1
 WD=$(pwd)
 
-echo "Rendering templates"
-bash ${WD}/test/config/render-ctmpls.sh "dev" "${VAULT_TOKEN}"
+
+INPUTS_JSON="/working/test/${PIPELINE_FOLDER_NAME}/pr/test_inputs.json"
```
ambrosejcarr: This is good for now. Longer term, I'd **strongly** suggest we make this structure a bit less restrictive; it's annoying (as perhaps you've just discovered!) to realize you need to name all the files _just so_.

test/test_cromwell_workflow.sh
jessicaway: That spelling error has been corrected, but there are still some bugs I will change this from draft once everything is working properly

test/trigger_test.sh
jessicaway: We can make this a parameter for the test if that is preferred. This file structure / naming convention is used in the portability test so we thought we would just take advantage of the required naming that is already in place to minimize parameters here

test/test_cromwell_workflow.sh
ambrosejcarr: Whoa, I have never seen draft PRs before. Sorry for pulling the trigger too early on the review!

test/trigger_test.sh
ambrosejcarr: On one hand, the UX of this is awful, since it's not very discoverable for users trying to write new tests (look at the commit flow for #184 , you can see that I wasn't able to guess right the first time). 

However, I'm also aware that it's used for all the workflow tests, so the scope of fixing this is much larger. Ideally, tests added to the testing folder would be automatically discovered by both the GCP and AWS/Local CI tests. I'd suggest filing away an issue for later?

jessicaway:
retest this please

test/trigger_test.sh
jessicaway: Yes, automatically discovered tests would be much better. I can add a ticket to make that happen in the future

test/test_cromwell_workflow.sh
```diff
@@ -2,25 +2,26 @@
...
 WORKFLOW_HASH=$(cromwell-tools submit \
-  --username "${CROMWELL_USER}" \
-  --password "${CROMWELL_PASSWORD}" \
+  --service-account-key "${CROMWELL_KEY_FILE}" \
   --url "${CROMWELL_URL}" \
-  --inputs_files "${INPUTS_JSON}" \
+  --inputs-files "${INPUTS_JSON}" \
   --wdl-file "${WDL_FILE}" \
   --options-file "${OPTIONS_FILE}" \
+  --collection-name ${COLLECTION} \
```
tbl3rd: Not quoted?

test/test_cromwell_workflow.sh
```diff
@@ -2,25 +2,26 @@
...
   --deps-file ${DEPENDENCIES_LIST[@]} \
```
tbl3rd: Is DEPENDENCIES_LIST an array. It doesn't look like one on line 13. Quote it?

test/trigger_test.sh
```diff
@@ -4,34 +4,35 @@ set -e
...
+CROMWELL_KEY_FILE="caas-prod.json"
 
 OPTIONS_FILE="https://raw.githubusercontent.com/HumanCellAtlas/skylab/master/test/options.json"
-CROMWELL_URL="https://cromwell.mint-dev.broadinstitute.org"
+CROMWELL_URL="https://cromwell.caas-prod.broadinstitute.org"
+COLLECTION="lira-dev"
 
 echo "Running test"
 docker run --rm \
   -v ${WD}:/working \
```
tbl3rd: Quote this too?

test/test_cromwell_workflow.sh
```diff
@@ -2,25 +2,26 @@
...
-DEPENDENCIES_JSON=$7
+CROMWELL_KEY_FILE=$1
+CROMWELL_URL=$2
+INPUTS_JSON=$3
+WDL_FILE=$4
+OPTIONS_FILE=$5
+DEPENDENCIES_JSON=$6
+COLLECTION=$7
 # Read list of dependency files into an array to pass to cromwell-tools
 DEPENDENCIES_LIST=$(cat ${DEPENDENCIES_JSON} | python3 -c "import json,sys;obj=json.load(sys.stdin);print(' '.join(obj.values()));")
```
tbl3rd: Could probably do this with jq too.

test/test_cromwell_workflow.sh
jessicaway: added

test/trigger_test.sh
jessicaway: added

test/test_cromwell_workflow.sh
jessicaway: @tbl3rd are you ok with this as is, or would you like to see a change?

PR #197, opened 2019-05-13
Add single-end Smart-Seq2 pipeline

mckinsel:
### Purpose
This tweaks the paired-end SS2 pipeline to work on single end data.

### Changes
The main change is just making `hisat` accept one fastq file instead of two. There's also a change in how `crimson` parses picard alignment summary results for unpaired reads that required a small change to sctools.


mckinsel:
retest

rexwangcc:
retest this please

library/tasks/GroupMetricsOutputs.wdl
```diff
@@ -1,15 +1,15 @@
...
   File hisat2_stats
   File hisat2_trans_stats
   File rsem_stats
   String output_name
   # Runtime
-  String docker = "quay.io/humancellatlas/secondary-analysis-sctools:v0.3.0"
-  String mem = 2
-  String cpu = 1
-  String disk_space = 20 
+  String docker = "quay.io/humancellatlas/secondary-analysis-sctools:mk-unpaired-reads"
```
barkasn: Change this to versioned

library/tasks/HISAT2.wdl
```diff
@@ -211,6 +211,7 @@ task HISAT2SingleEnd {
   # Need room for unsorted + sorted bam + temp sorting space + zipped and unzipped ref. Add 10 GB buffer.
   Int disk = ceil((size(fastq, "GB") * 100) + size(hisat2_ref, "GB") * 2 + 10)
```
barkasn: Confirm with @kbergin if these should be GiBs or not yet

library/tasks/HISAT2.wdl
```diff
@@ -240,8 +241,11 @@ task HISAT2SingleEnd {
       --new-summary --summary-file "${output_basename}.log" \
       --met-file ${output_basename}.hisat2.met.txt --met 5 \
       --seed 12345 \
+      -k 10 \
+      --secondary \
```
barkasn: Why add secondary only in single end? I would argue for keeping this consistent with pe.

test/smartseq2_single_sample_unpaired/pr/ValidateSmartSeq2SingleCell.wdl
```diff
@@ -0,0 +1,43 @@
...
+      >&2 echo "counts_hash ($gene_counts_hash) did not match expected hash (${expected_gene_counts_hash})"
+      fail=true
+    fi
+
+    if [ $fail == "true" ]; then exit 1; fi
+
+  >>>
+  
+  runtime {
+    docker: "ubuntu:16.04"
```
barkasn: Consider using latest ubuntu

test/smartseq2_single_sample_unpaired/pr/ValidateSmartSeq2SingleCell.wdl
```diff
@@ -0,0 +1,43 @@
...
+    cpu: 1
+    memory: "3.75 GB"
```
barkasn: GiB?

library/tasks/HISAT2.wdl
mckinsel: Paired end does use secondary. These parameters were added to the hisat call matches the pe call. I think this task was unused and had diverged from the pe task.

test/smartseq2_single_sample_unpaired/pr/ValidateSmartSeq2SingleCell.wdl
mckinsel: As far as I can tell, we always use "GB" in all our tasks and never "GiB"

test/smartseq2_single_sample_unpaired/pr/ValidateSmartSeq2SingleCell.wdl
mckinsel: Updated

library/tasks/HISAT2.wdl
mckinsel: According to [the spec](https://github.com/openwdl/wdl/blob/master/versions/1.0/SPEC.md#float-sizefile-string), either is supported for the `size` call, but it looks like [Cromwell expects](https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#disks) that number to mean GB.

library/tasks/GroupMetricsOutputs.wdl
mckinsel: Okay will do once the new version of sctools is released.

test/smartseq2_single_sample_unpaired/pr/ValidateSmartSeq2SingleCell.wdl
rexwangcc: GiB is a new thing in Google Pipelines API v2, fyi, https://github.com/HumanCellAtlas/secondary-analysis/issues/446. We have updated to use it in some Optimus dependency WDLs, such as https://github.com/HumanCellAtlas/skylab/blob/bfd9051abdfd5aca7a6e92248d1b55ccfb737f52/library/tasks/Attach10xBarcodes.wdl#L12 but haven't updated to use GiB in our adapter WDLs yet unfortunately.

library/tasks/HISAT2.wdl
ambrosejcarr: A single PR should update `GB` -> `GiB` if/when you decide to make the change; until then users should be expected to pattern match on existing code as Marcus has done.

library/tasks/HISAT2.wdl
```diff
@@ -240,8 +241,11 @@ task HISAT2SingleEnd {
       --new-summary --summary-file "${output_basename}.log" \
       --met-file ${output_basename}.hisat2.met.txt --met 5 \
       --seed 12345 \
+      -k 10 \
```
ambrosejcarr: Is it possible to record the logic behind setting these parameters? I know the existing WDL sets a bad precedent, but it's nice to be able to understand why, e.g. `k=10`.

library/tasks/HISAT2.wdl
```diff
@@ -304,3 +310,94 @@ task HISAT2InspectIndex {
...
+    fi
+
+
+    tar --no-same-owner -xvf "${hisat2_ref}"
+
+    # increase gap alignment penalty to avoid gap alignment
+    # --mp 1,1 --np 1 --score-min L,0,-0.1 is default paramesters when rsem runs alignment by using bowtie2/Bowtie
+    # --mp 1,1 and --np 1 will reduce mismatching penalty to 1 for all.
+    # with no-splice-alignment no-softclip no-mixed options on, HISAT2 will only output concordant alignment without soft-cliping
+    # --rdg 99999999,99999999 and --rfg 99999999,99999999 will give an infinity penalty to alignment with indel.As results
```
ambrosejcarr:
```suggestion
    # --rdg 99999999,99999999 and --rfg 99999999,99999999 set an infinite penalty to alignments with indels. 
    # As a result, alignments with gaps or deletions are excluded. 
```
Fixed a copy/pasted typo, ended up rewriting the sentence to make sense.

mckinsel: Thanks fixed.

library/tasks/HISAT2.wdl
mckinsel: So there is a comment in the WDL above, and it turns out `-k 10` is just the default behavior or HISAT2. I can at least explain that this is copied from the PE task.

library/tasks/HISAT2.wdl
samanehsan: There was a change in Cromwell recently so that the size function returns GiB even if you ask for GB so we should definitely change this to use GiB to be explicit: https://github.com/broadinstitute/cromwell/issues/4906

pipelines/smartseq2_single_sample_unpaired/SmartSeq2UnpairedExample.json
```diff
@@ -0,0 +1,15 @@
+{
+  "SmartSeq2SingleCell.hisat2_ref_trans_name": "gencode_v27_trans_rsem",
```
samanehsan: This should be `SmartSeq2SingleCell` -> `SmartSeq2SingleCellUnpaired`

.circleci/config.yml
```diff
@@ -10,7 +10,9 @@ jobs:
                     python3 -m venv venv
                     . venv/bin/activate
                     pip install requests
-                    python test/run_portability_test.py --test-directories test/smartseq2_single_sample/pr/ test/optimus/pr
+                    python test/run_portability_test.py --test-directories test/smartseq2_single_sample/pr/ \
+                                                                           test/smartseq2_single_sample_unpaired/pr \
+                                                                           test/optimus/pr
```
pshapiro4broad: Can you also add a trigger_test version of this test below? Once you update from master, there will be an example of how to do that.

pipelines/smartseq2_single_sample_unpaired/SmartSeq2UnpairedExample.json
mckinsel: Thanks, fixed

library/tasks/HISAT2.wdl
mckinsel: Okay, updated to `GiB`

.circleci/config.yml
mckinsel: Okay, rebased and added a trigger_test to circleci

PR #209, opened 2019-05-15
Optimus 10X V3 chemistry

mlin:
To be squashed with commit message:

Optimus has a new optional input tenX_v3_chemistry, when set to true it reconfigures the AttachBarcodes step to extract 12bp UMI barcodes instead of the default 10bp assumed for v2 chemistry. This functionality is tested with small FASTQs excerpted from a public 10X v3 dataset, staged in the cloud storage bucket.


test/optimus/v3umi/TestOptimusV3UMI.wdl
```diff
@@ -0,0 +1,80 @@
...
+        set -eux -o pipefail
+        apt-get update && apt-get install -y samtools
+        L="$(samtools view '${optimus_bam}' | grep -E -o 'UR:Z:[ACGTN]+' | awk '{print length}' | sort -u)"
+        if [ "$L" -ne "${that_plus5}" ]; then
+            exit 1
+        fi
+    >>>
+
+    runtime {
+        docker: "ubuntu:18.04"
```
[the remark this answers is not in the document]
benjamincarlin: either quay.io/humancellatlas/secondary-analysis-samtools or biocontainers/samtools should work

mlin:
@benjamincarlin @mckinsel I addressed the comments and this might be just about squash+mergeable depending on what CircleCI thinks. Thanks!

Optimus has a new optional input tenX_v3_chemistry, when set to true it reconfigures the AttachBarcodes step to extract 12bp UMI barcodes instead of the default 10bp assumed for v2 chemistry. This functionality is tested with small FASTQs excerpted from a public 10X v3 dataset, staged in the cloud storage bucket.

barkasn:
I propose we have a discussion before we merged this about the best way to parametrise the pipeline. I think that a single flag, that is not backwards compatible (i.e. changes default behaviour) is not the best way. I propose we have a text field that identified the input data variant and we potentially harmonize this with the other variants of data we are adding to the pipeline.

library/tasks/Attach10xBarcodes.wdl
```diff
@@ -3,9 +3,10 @@ task Attach10xBarcodes {
   File? i1_fastq
   File r2_unmapped_bam
   File whitelist
+  Boolean tenX_v3_chemistry = false
 
   # runtime values
-  String docker = "quay.io/humancellatlas/secondary-analysis-sctools:v0.3.2"
+  String docker = "quay.io/humancellatlas/secondary-analysis-sctools:v0.3.4"
```
barkasn: Needs to be updated

library/tasks/Attach10xBarcodes.wdl
```diff
@@ -21,7 +22,8 @@ task Attach10xBarcodes {
     r1_fastq: "forward fastq file; contains umi, cell barcode"
     i1_fastq: "optional, index fastq file; contains sample barcode"
     r2_unmapped_bam: "reverse unmapped bam file; contains alignable genomic information"
-    whitelist: "10x genomics cell barcode whitelist for 10x V2"
+    whitelist: "10x genomics cell barcode whitelist"
+    tenX_v3_chemistry: "assume 10X Genomics v3 chemistry with 12bp UMI (in contrast to default v2 with 10bp UMI)"
```
barkasn: Consider re-parametrization

library/tasks/Attach10xBarcodes.wdl
```diff
@@ -32,12 +34,22 @@ task Attach10xBarcodes {
   command {
     set -e
 
-    Attach10xBarcodes \
```
barkasn: Consider taking the if out one level, this looks messy

pipelines/optimus/Optimus.wdl
```diff
@@ -32,8 +32,9 @@ workflow Optimus {
   File annotations_gtf
   File ref_genome_fasta
 
-  # 10x v2 parameters
+  # 10x parameters
   File whitelist
+  Boolean tenX_v3_chemistry = false
```
barkasn: re-parametrize

test/optimus/v3umi/TestOptimusV3UMI_PR.wdl
```diff
@@ -0,0 +1,86 @@
+# test extraction of 10X Genomics v3 chemistry's 12bp UMI barcodes
```
barkasn: Requires updating with data from chr21

test/optimus/v3umi/test_inputs.json
```diff
@@ -0,0 +1,16 @@
+{
```
barkasn: Requires updating

library/tasks/Attach10xBarcodes.wdl
```diff
@@ -32,22 +34,45 @@ task Attach10xBarcodes {
...
-      --r1 "${r1_fastq}" \
-      ${"--i1 " + i1_fastq} \
-      --u2 "${r2_unmapped_bam}" \
-      --output-bamfile barcoded.bam \
-      --whitelist "${whitelist}"
+    if [ "${chemistry}" == "tenX_v2" ]
```
tlangs: It might be cleaner to have different tasks for the different chemistries and run them using an `if` condition at the workflow level instead of the task level. That would mean a task always produces the "same" types of outputs and it's clear what was run just by looking at a timing diagram rather than having to drill into the metadata to get inputs at runtime.

barkasn: Would normally agree but eventually we only want one command 'AttachBarcodes' with chemistry dependent parameters (that are ideally read from a file not from he command line). We are keeping Attach10xBarcodes because we don't want to alter the code path for V2 and risk breaking things. The whole thing needs to be revised anyway because its inefficient and currently the slowest pipeline step.

tlangs: ah ok

barkasn: @kbergin tagging you to be aware for future prioritization

mlin:
@barkasn @HumanCellAtlas/pipelines-computational-biologists can anything be done to cut down the 28GB star_primary_gencode_v27.tar now used in the PR tests? It's a bummer for anyone trying to test out the workflow locally to have to download that. Asking for a friend

ambrosejcarr:
Previously, we used a reference with only chromosome 19 and 21 which was much smaller.

PR #189, opened 2019-04-30
GL-132 increase memory in optimus and ss2

jessicaway:
### Purpose
_Please explain the purpose of this PR and include links to any GitHub issues that it fixes:_
Changed MB/GB to MiB/GiB to reflect the way memory is allocated in Cromwell 38 and PAPIv2. Increased the memory size by 10% to adjust for this change.
- No issue is linked to this PR.
---
### Changes
_Please list out what major changes were made in this PR to address the issue:_

- No changes.
---
### Review Instructions
_Please provide instructions about how should a reviewer test/verify the changes in this PR:_

- No instructions.

---
### PR Checklist
_Please ensure the following when opening a PR:_

- [ ] This PR added or updated tests.
- [ ] This PR updated docstrings or documentation.
- [ ] Let Jodi (**jodihir**) know when there is user-facing documentation updates needed!
- [ ] This PR applied the Mint style guide for WDL.
- [ ] This PR has no WDL linting errors reported by [miniwdl](https://github.com/chanzuckerberg/miniwdl)
- [ ] This PR considered generalizability beyond our own use case.

---
### Follow-up Discussions
_Please append follow-up discussions and issues during the review process below:_

- No follow-up discussions.

PR #215, opened 2019-05-22
Improve Performance of SplitBamByCellBarcode

tlangs:
### Purpose

`SplitBam` is slow. We want to make it faster.

### Changes

Completely rewrote `SplitBam` to take advantage of parallelization. The tool now takes a list of bams as input: `SplitBam -b bam1.bam bam2.bam` as well as an optional `--num-threads` to set the number of threads/cores to use. 

Testing with the bam produced by the 4k_pbmc dataset, `SplitBam` task took 50 minutes using 8 threads. For reference, the old implementation took 217 minutes. Both of these tasks were run using HDDs, so we may see even more speedup when using SSDs in the parallelized implementation.

### Review Instructions

- No instructions.


tlangs:
The bam_hash changed because the new implementation orders tags slightly differently than the old version.

library/tasks/ScatterBam.wdl
```diff
@@ -0,0 +1,27 @@
...
+    Array[File] scattered_bams = glob("scattered_bams/*")
+  }
+
+  runtime {
+    disk: "local-disk ${disk_size} HDD"
+    cpu: 2
+    memory: "7.5 GB"
+    docker: "us.gcr.io/broad-gotc-prod/genomes-in-the-cloud:2.3.3-1513176735"
+  }
+}
```
rexwangcc:
```suggestion

```
(nit)

rexwangcc:
```suggestion
}

```

library/tasks/ScatterBam.wdl
```diff
@@ -0,0 +1,27 @@
...
+  >>>
+
+  output {
+    Array[File] scattered_bams = glob("scattered_bams/*")
+  }
+
+  runtime {
+    disk: "local-disk ${disk_size} HDD"
+    cpu: 2
+    memory: "7.5 GB"
```
barkasn: Should this be GiB or not yet?

test/optimus/pr/test_inputs.json
```diff
@@ -1,5 +1,5 @@
 {
-  "TestOptimusPR.expected_bam_hash": "a54495ad22db2dd9b08b84b999fdb137",
+  "TestOptimusPR.expected_bam_hash": "afd10dcbdb0055c90b41c4279d11eeae",
```
barkasn: Ok following our discussion about order of tags

library/tasks/ScatterBam.wdl
ambrosejcarr: Personal preference: style should always be consistent across the codebase. Right now we use `GB`. If we want or need to change to `GiB`, that should be done as a single refactor that updates the entire codebase of skylab, not in individual PRs (which would cause us to have both `GB` and `GiB` simultaneously).

tlangs: Is Optimus running on PAPIv2?

rexwangcc: Yes, I think all HCA pipelines are running on PAPIv2: https://github.com/HumanCellAtlas/pipeline-tools/blob/master/adapter_pipelines/Optimus/options.json

tlangs: Then it should be GiB. I'll change it.

ambrosejcarr: Do we have any pipelines that are run on PAPIv1?

If not, I would prefer someone to go through our codebase and make the relevant `s/GB/GiB` refactor. It is confusing to have two styles present simultaneously.

jessicaway: We changed all the tasks used by SS2 and Optimus to use GiB in [this PR](https://github.com/HumanCellAtlas/skylab/pull/189) to fix out of memory issues. We can put in a ticket to change the rest of the tasks in Skylab

PR #217, opened 2019-06-03
Increase disk size for TagGeneExon

jessicaway:
### Purpose
We were seeing some issues with the TagGeneExon running out fo Disk space in Optimus. This PR increases the disk space from for the TagGeneExon task. 

### Changes
- Increase disk size in TagGeneExon task

### Review Instructions

- No instructions.


samanehsan:
After this is merged we should tag a new version of the Optimus pipeline!

PR #211, opened 2019-05-17
Add WDLized form of SNAP-ATAC Pipeline

barkasn:
### Purpose
The purpose of this PR is to add the [snap-atac pipeline primary processing pipeline](https://github.com/r3fang/SnapATAC) to the repository. 

### Changes
* Dockerfile for building image
* WDL for running the pipeline
* WDL for building BWA pipeline
* Tests for pipeline

### Review Instructions

- No instructions.

PR #219, opened 2019-06-03
Improve portability test log access

mckinsel:
This is a small change to the portability test script to format the logs nicely and print them even for runs that succeed.


test/run_portability_test.py
```diff
@@ -177,11 +177,10 @@ def monitor_tests(test_ids):
 
     for test_name, test_id in test_ids.items():
 
-        if terminal_tests[test_name] != "Succeeded":
```
pshapiro4broad: Is there a benefit to producing output if the build passes? It seems to me that this adds unnecessary noise to the build output.

mckinsel: It was a user request. I think there are cases where the job technically succeeds, but the developer would still like to sniff around the logs to see what exactly happened.

mckinsel:
These changes were included in #211

PR #226, opened 2019-06-27
GH-282: move max retries out of wdls

ApChagi:
### Purpose

Ticket: [JIRA](https://broadinstitute.atlassian.net/browse/GH-282)

Instead of passing in the maxRetries runtime parameter into the workflow, and then passing that into every task, define it in the workflow options file.

### Changes

- maxRetries runtime parameter moved out of wdl & task inputs and into cromwell options by way of Lira config
[parallel changes in lira](https://github.com/HumanCellAtlas/lira/pull/168)
[parallel changes in pipeline-tools](https://github.com/HumanCellAtlas/pipeline-tools/pull/151)

Note: This change needs to be merged in at the same time as the pipeline-tools change, otherwise the mintegration tests that run off of our PRs will break.

### Review Instructions

- No instructions.

### DIFF
--- a/pipelines/cellranger/cellranger.wdl
+++ b/pipelines/cellranger/cellranger.wdl
@@ -16,7 +16,6 @@ workflow CellRanger {
     Int boot_disk_size_gb = 12
     String disk_space = "400"
     Int cpu = 64
-    Int max_retries = 0
 
     parameter_meta {
         sample_id: "Name of sample to run CellRanger count on"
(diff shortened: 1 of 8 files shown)


pipelines/cellranger/cellranger.wdl
```diff
@@ -117,7 +113,6 @@ task cellranger_count {
         bootDiskSizeGb: boot_disk_size_gb
         disks: "local-disk " + disk_space + " HDD"
         cpu: cpu
-        max_retries: max_retries
```
pshapiro4broad: As it turns out, all the uses of `max_retries` here are bogus, as the runtime option is spelled `maxRetries`, so the original code had no effect.

samanehsan:
The optimus test failed because of the time-out limit, but the test workflow succeeded: https://job-manager.caas-prod.broadinstitute.org/jobs/4c1c56cf-0b31-4d3d-98e0-d67966e0695f

PR #228, opened 2019-07-08
GL-407 Fix OODisk error when creating count matrix

infispiel:
### Purpose
We are encountering an issue in Optimus production where due to a change where we have started modifying gtfs not in-place, and due to this we are running out of hard drive space on the create count matrix task. This PR is to change the way we modify the gtf to both make it more DRY and to no longer cause this error.

[Jira Ticket [GL-407]](https://broadinstitute.atlassian.net/browse/GL-407)

### Changes
- Created separate task to modify GTF
- Moved python logic into python script with a dedicated docker
- Changed code to reference outputs to separate task rather than re-doing work

### Review Instructions
Run Optimus on a large dataset. This previously passed review due to the fact the size only became an issue with large-ish GTFs.

### DIFF
--- a/docker/modify-gtf/SetGeneNameToId.py
+++ b/docker/modify-gtf/SetGeneNameToId.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+
+import argparse
+import re
+
+parser = argparse.ArgumentParser(description="Set the gene_name within a gtf to be equivalent to the values within gene_id.")
+parser.add_argument('--in-gtf-file', dest='ingtf', help='input gtf file')
+parser.add_argument('--out-gtf-file', dest='outgtf', help='output gtf file')
+
+args = parser.parse_args()
+
+
+def setGeneNameToId(in_gtf, out_gtf, verbose=True):
+    with open(in_gtf, 'r') as fpin, open(out_gtf, 'w') as fout:
+        for _line in fpin:
+            line = _line.strip()
+            gene_id = re.search(r'gene_id ([^;]*);', line)
+            gene_name = re.search(r'gene_name ([^;]*);', line)
+            if gene_id and gene_name:
+                modified_line = re.sub(r'gene_name ([^;]*);', 'gene_name ' + gene_id.group(1) + ";", line)
+                fout.write(modified_line + '\n')
+            else:
+                fout.write(line + '\n')
+
+setGeneNameToId(args.ingtf,args.outgtf)
--- a/docker/modify-gtf/SetGeneNameToId.py
+++ b/docker/modify-gtf/SetGeneNameToId.py
@@ -11,15 +11,15 @@ args = parser.parse_args()
 
 
 def setGeneNameToId(in_gtf, out_gtf, verbose=True):
-    with open(in_gtf, 'r') as fpin, open(out_gtf, 'w') as fout:
-        for _line in fpin:
-            line = _line.strip()
-            gene_id = re.search(r'gene_id ([^;]*);', line)
-            gene_name = re.search(r'gene_name ([^;]*);', line)
+    with open(in_gtf, 'r') as fpin, open(out_gtf, 'w') as fpout:
+        for line in fpin:
+            stripped_line = line.strip()
+            gene_id = re.search(r'gene_id ([^;]*);', stripped_line)
+            gene_name = re.search(r'gene_name ([^;]*);', stripped_line)
             if gene_id and gene_name:
-                modified_line = re.sub(r'gene_name ([^;]*);', 'gene_name ' + gene_id.group(1) + ";", line)
-                fout.write(modified_line + '\n')
+                modified_line = re.sub(r'gene_name ([^;]*);', 'gene_name ' + gene_id.group(1) + ";", stripped_line)
+                fpout.write(modified_line + '\n')
             else:
-                fout.write(line + '\n')
+                fpout.write(stripped_line + '\n')
 
 setGeneNameToId(args.ingtf,args.outgtf)
--- a/library/tasks/ModifyGtf.wdl
+++ b/library/tasks/ModifyGtf.wdl
@@ -3,7 +3,8 @@ task ReplaceGeneNameWithGeneID {
   String modified_gtf_location = "gene_id_as_gene_name.gtf"
 
   #runtime values
-  String docker = "quay.io/humancellatlas/modify-gtf:0.1.0"
+  #String docker = "quay.io/humancellatlas/modify-gtf:0.1.0"
+  String docker = "us.gcr.io/broad-gotc-dev/shpilker_modify-gtf"
   Int machine_mem_mb = 8250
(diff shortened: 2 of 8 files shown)


pshapiro4broad:
The phrase `modify gtf` seems too general to me, is there a more specific name for what this code is doing? `ReplaceGeneNameWithGeneID` is pretty long but at least it's clear what it's doing, or maybe `UpdateGtfWithGeneId`? This would be for the WDL, the directory, docker image name, etc.

infispiel:
That was my initial plan, but then I started looking through the other tasks files and saw that they had multiple tasks that had more specialized names. I then opted to name it "modify gtf" in case we will add more functions in the future that also modify the gtf & therefore we can wrap them inside of the same task file, docker, etc; this may be thinking ahead too far, however, as this is a temporary bandaid. What is your opinion on those comments?

library/tasks/ModifyGtf.wdl
```diff
@@ -0,0 +1,51 @@
+task ReplaceGeneNameWithGeneID {
```
ambrosejcarr: Great changes. Can you comment on why we don't we take this one step further?

1. execute this script one time
2. save it somewhere
3. pass in the modified GTF file to the Optimus script

docker/modify-gtf/SetGeneNameToId.py
```diff
@@ -0,0 +1,25 @@
...
+            line = _line.strip()
+            gene_id = re.search(r'gene_id ([^;]*);', line)
+            gene_name = re.search(r'gene_name ([^;]*);', line)
+            if gene_id and gene_name:
+                modified_line = re.sub(r'gene_name ([^;]*);', 'gene_name ' + gene_id.group(1) + ";", line)
+                fout.write(modified_line + '\n')
+            else:
+                fout.write(line + '\n')
+
+setGeneNameToId(args.ingtf,args.outgtf)
```
ambrosejcarr:
```suggestion
setGeneNameToId(args.ingtf, args.outgtf)
```

linting ;-)

library/tasks/ModifyGtf.wdl
infispiel: Thank you for your feedback, Ambrose! I think while it seems nice to not have to re-modify the gtf each time we run the workflow, I think there are some things that make it sub-optimal to preemptively create & save the file:

1. That would involve a change to our infrastructure rather than a workflow, and this approach isn't set in stone. We are still testing whether this is the best way to modify the gtf, and if the 3rd party tools we create this for get changed/updated we may no longer need this file.

2. From what I understand from discussions with Comp Bios, we don't want the modified gtf to be an output or otherwise accessible since it's technically a workaround and a big no-no.

3. The gtf is very small (comparatively) and the operation to modify it fast enough that it isn't really an issue to do. Only 2 tasks need the output from ReplaceGeneNameWithGeneID, so hopefully there is time for it to run in parallel with other tasks. Furthermore, if call-cacheing is enabled, it will immediately return a cache hit.

docker/modify-gtf/SetGeneNameToId.py
```diff
@@ -0,0 +1,25 @@
...
+
+parser = argparse.ArgumentParser(description="Set the gene_name within a gtf to be equivalent to the values within gene_id.")
+parser.add_argument('--in-gtf-file', dest='ingtf', help='input gtf file')
+parser.add_argument('--out-gtf-file', dest='outgtf', help='output gtf file')
+
+args = parser.parse_args()
+
+
+def setGeneNameToId(in_gtf, out_gtf, verbose=True):
+    with open(in_gtf, 'r') as fpin, open(out_gtf, 'w') as fout:
```
tlangs: `fpin` to me means `file pointer in`, so shouldn't `fout` be `fpout`?

library/tasks/ModifyGtf.wdl
```diff
@@ -0,0 +1,51 @@
...
+  File original_gtf
+
+  #runtime values
+  String docker = "quay.io/humancellatlas/modify-gtf:0.1.0"
+  Int machine_mem_mb = 8250
+  Int cpu = 1
+  Int disk = ceil(size(original_gtf, "Gi") * 2) + 1
```
tlangs: What unit is `Gi`? I've only seen `GB` and `GiB`

library/tasks/ModifyGtf.wdl
```diff
@@ -0,0 +1,51 @@
...
+    cpu: "(optional) the number of cpus to provision for this task"
+    disk: "(optional) the amount of disk space (GiB) to provision for this task"
+    preemptible: "(optional) if non-zero, request a pre-emptible instance and allow for this number of preemptions before running the task on a non preemptible machine"
+  }
+
+  command {
+    set -e
+
+    if file --mime-type "${original_gtf}" | grep  gzip; then
+      gunzip -c  "${original_gtf}" > input.gtf
```
tlangs: Why are we supporting both zipped and non-zipped gtf files? Can we choose one to support and stick with it?

library/tasks/ModifyGtf.wdl
infispiel: All of Optimus actually uses the `Gi` syntax, so I would rather leave it consistent to be fixed in a different PR. I share the same confusion.

library/tasks/ModifyGtf.wdl
infispiel: I just spoke to Hornet and it seems like they use externally-provided gtf files, and there isn't always certainty whether or not they come compressed or not, hence the logic to unzip if necessary.

docker/modify-gtf/SetGeneNameToId.py
infispiel: Changed!

docker/modify-gtf/SetGeneNameToId.py
```diff
@@ -0,0 +1,25 @@
...
+        for _line in fpin:
+            line = _line.strip()
+            gene_id = re.search(r'gene_id ([^;]*);', line)
+            gene_name = re.search(r'gene_name ([^;]*);', line)
```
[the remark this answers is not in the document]
infispiel: Changed!

library/tasks/ModifyGtf.wdl
tlangs: Shouldn't the adapter workflows take care of providing the workflow consistent data? Lets not let it get in the way of this PR, but its something we should keep in mind.

library/tasks/ModifyGtf.wdl
tlangs: That's fair.

library/tasks/ModifyGtf.wdl
kbergin: Note that we also develop these pipelines for users, who may not be as consistent. But I agree, it's reasonable to set standards or define inputs as long as we communicate the pipeline's expectations clearly in our documentation.

PR #231, opened 2019-07-18
HISAT2: fix the shell brackets used for filename extension tests

mlin:
These task commands attempt to detect if the input filenames end in .fastq.gz, and if not then append this extension. But they're using the wrong kind of shell brackets for this extension test, resulting in the renaming occurring unconditionally.

```
$ if [ "foo.fastq.gz" != *.fastq.gz ]; then echo 1; else echo 0; fi
1
$ if [[ "foo.fastq.gz" != *.fastq.gz ]]; then echo 1; else echo 0; fi
0
```


kbergin:
Hi Mike,

We are moving repositories and have already moved the code impacted here, so [I have replicated the changes in this PR](https://github.com/broadinstitute/warp/pull/16) in a branch there, which is testing now. I am going to close this PR. 

Hope you are well

PR #234, opened 2019-07-22
GL-416: Improvements to Optimus Testing

barkasn:
### Purpose
This PR is addressing the following ticket
https://broadinstitute.atlassian.net/jira/software/projects/GL/boards/528?selectedIssue=GL-416

### Changes
- Remove Scientific Tests that are out of date and are not being used
- Separate the portability tests into distinct circleci tasks to allow identifying failures more easily
- Replace the dataset used to do the testing with something that gives meaningful matrix output
- Harmonize the reference with the integration tests
- Add instructions for running tests manually to README
- Convert Optimus Testing Task into a workflow and split out tasks
- Compare the Optimus BAM Output with previous outputs only looking at primary reads and removing extra tags to allow for UMI correction stochasticity
- Compare Optimus Matrix with a reference matrix, instead of checksum that is dependent on compression stochasticity. New comparison also ignores changes in the order of cells.
- Auto-generate new reference matrix files on each run to ease of test updates
- Generate three simple QC plots on matrix to help diagnose problems
- Update .gitignore to avoid accidental credential check in and emacs backup files
- Update test README.txt
- Remove dummy build step from circleci
- Increase timeout of tests to 90 to avoid failing Optimus tests without cache

### Review Instructions
- Check code quality
- The npz2rds utility is now used in two different images (emptyDrops and now in optimus-matrix-testing). Should we factor this out as a separate repository? (I think yes, and if so this should be another ticket)

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -1,3 +1,15 @@
+**Running Tests Manually**
+1) Set the environment variable BROAD_CROMWELL_KEY to have the contents of the cromwell credentials key JSON file.
+2) cd to the top level directory of the repository
+3) run the following command ```./test/trigger_test.sh [name of workflow without trailing /pr/]```
+
+Example:
+```
+export BROAD_CROMWELL_KEY=`cat ~/identities/credentials.json`
+cd ~/skylab/
+./test/trigger_test.sh optimus
+```
+
 **Scientific and PR testing infrastructure in Skylab.**
 
 Both tests share the same infrastructure, and follow the portability spec; the WDLs in this PR represent a cromwell test case for the [portability test](https://docs.google.com/document/d/1ghLoHMbKOPsndA1WgdSAHm5X82p86ryLBiAt1hz6HuI/edit); they are just missing a docker environment to run the WDLs in. 
--- a/test/README.md
+++ b/test/README.md
@@ -10,48 +10,26 @@ cd ~/skylab/
 ./test/trigger_test.sh optimus
 ```
 
-**Scientific and PR testing infrastructure in Skylab.**
-
-Both tests share the same infrastructure, and follow the portability spec; the WDLs in this PR represent a cromwell test case for the [portability test](https://docs.google.com/document/d/1ghLoHMbKOPsndA1WgdSAHm5X82p86ryLBiAt1hz6HuI/edit); they are just missing a docker environment to run the WDLs in. 
+**PR testing infrastructure in Skylab.**
+The tests share follow the portability spec; the WDLs in this PR represent a cromwell test case for the [portability test](https://docs.google.com/document/d/1ghLoHMbKOPsndA1WgdSAHm5X82p86ryLBiAt1hz6HuI/edit); they are just missing a docker environment to run the WDLs in. 
 
 The PR test is designed to answer the question: "did my changes result in any change to the outputs of my pipeline?"
 
-The scientific test is designed to answer the question: "Is the pipeline performing within an expected range of scientific outputs?". 
-It compares the results of optimus to the results of the Cell Ranger comparison pipeline, run on the same data. 
-
 **Testing Logic & Files**
-All files added in this PR are contained within the test directory. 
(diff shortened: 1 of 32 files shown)


test/optimus/pr/ValidateOptimus.wdl
```diff
@@ -1,64 +1,223 @@
...
+        fail=true
     fi
 
     if [ $fail == "true" ]; then exit 1; fi
 
   >>>
   
   runtime {
-    docker: "quay.io/humancellatlas/secondary-analysis-samtools:v0.2.2-1.6"
+    docker: "ubuntu:18.04"
```
pshapiro4broad: We'd like to move to using google managed containers where possible (and it makes sense to do so). Would this work?
```suggestion
    docker: "marketplace.gcr.io/google/ubuntu1804"
```

barkasn: Yes for all stock ubuntu, for others it would be another ticket. I'll update.

test/README.md
```diff
@@ -1,45 +1,35 @@
...
+Example:
+```
+export BROAD_CROMWELL_KEY=`cat ~/identities/credentials.json`
+cd ~/skylab/
+./test/trigger_test.sh optimus
+```
 
-The PR test is designed to answer the question: "did my changes result in any change to the outputs of my pipeline?"
+**PR testing infrastructure in Skylab.**
+The tests share follow the portability spec; the WDLs in this PR represent a cromwell test case for the [portability test](https://docs.google.com/document/d/1ghLoHMbKOPsndA1WgdSAHm5X82p86ryLBiAt1hz6HuI/edit); they are just missing a docker environment to run the WDLs in.
```
pshapiro4broad: Not sure about this text
> The tests share follow the portability spec

test/optimus/pr/ValidateOptimus.wdl
```diff
@@ -1,64 +1,223 @@
...
-    cell_metric_hash=$(zcat "${cell_metrics}" | md5sum | awk '{print $1}')
+    command <<<
+        echo Starting checksum generation...
+        # calculate hash for alignment positions only (a reduced bam hash)
+        samtools view -F 256 "${bam}" | cut -f 1-11 | md5sum | awk '{print $1}' > md5_checksum_reduced.txt
+        echo Reduced checksum generation complete
 
-    # calculate hash as above, but ignore run-specific bam headers
-    bam_hash=$(samtools view "${bam}" | md5sum | awk '{print $1}')
+        calculated_checksum=$(cat md5_checksum_reduced.txt)
```
pshapiro4broad: You could avoid the file and write directly to the environment variable.
```
calculated_checksum=$(samtools view -F ... | awk '{print $1}')
```

test/optimus/pr/ValidateOptimus.wdl
```diff
@@ -1,64 +1,223 @@
...
+  
+    runtime {
+        docker: "quay.io/humancellatlas/secondary-analysis-samtools:v0.2.2-1.6"
+        cpu: 1
+        memory: "3.75 GB"
+        disks: "local-disk ${required_disk} HDD"
+    }
+
+    output {
+        String result = read_string("result.txt")
```
pshapiro4broad: Did you consider `Boolean result = read_boolean("result.txt")`?

test/optimus/pr/ValidateOptimus.wdl
barkasn: done

test/README.md
barkasn: fixed grammar, do you think the content needs revision too?

test/optimus/pr/ValidateOptimus.wdl
barkasn: Yes but I get a bit confused with true in bash due to inversion of regular 1/0 scemantics and I wasn't sure what Boolean was here so avoided it. Also thought String can potentially pass other statuses, e.g. warning better. Happy to change but will need a bit of help

PR #235, opened 2019-07-23
Enable subsetting by BED file

barkasn:
Allows subsetting fastq files by interval bed instead of a parameter to the CLI. This allows subsetting for gene regions for example.


barkasn:
This should be ready to merge, but as it doesn't have tests, I am waiting for the example workflow I am running to complete before requesting Eng review. The workflow can be found here:

https://job-manager.caas-prod.broadinstitute.org/jobs/0883bb08-3504-482a-94cb-670e1069aadc

PR #239, opened 2019-07-25
Set ScatterBam docker to be a public Picard docker.

infispiel:
### Purpose
- Some users are unable to use the Optimus pipeline due to docker image restrictions.
- We shouldn't be using the docker image used previously in ScatterBam.
- [Related Issue: GL-473](https://broadinstitute.atlassian.net/browse/GL-473)

### Changes
- Update picard Dockerfile to represent new maintainer & version.
- Change ScatterBam to use a just-created quay.io picard image.

### Review Instructions
- Run Optimus, and get the same-enough results as using the old ScatterBam docker image.

### DIFF
--- a/library/tasks/ScatterBam.wdl
+++ b/library/tasks/ScatterBam.wdl
@@ -1,3 +1,4 @@
+
 task ScatterBam {
 
   File bam_to_scatter
@@ -6,7 +7,7 @@ task ScatterBam {
 
   command <<<
     mkdir scattered_bams
-    java -Xms7g -jar /usr/gitc/picard.jar \
+    java -Xms7g -jar /usr/picard/picard.jar \
       SplitSamByNumberOfReads \
       INPUT=${bam_to_scatter} \
       SPLIT_TO_N_FILES=${scatter_width} \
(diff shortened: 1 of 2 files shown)


library/tasks/ScatterBam.wdl
```diff
@@ -22,6 +22,6 @@ task ScatterBam {
     disks: "local-disk ${disk_size} HDD"
     cpu: 2
     memory: "7.5 GiB"
-    docker: "us.gcr.io/broad-gotc-prod/genomes-in-the-cloud:2.3.3-1513176735"
+    docker: "quay.io/humancellatlas/secondary-analysis-picard:2.20.4"
```
pshapiro4broad: Is there a reason to prefer the quay.io one over the dockerhub one?

infispiel: I picked quay to be consistent with other docker images, since it sounds like HCA mostly pushes to quay. Does dockerhub have a recent picard docker image? I had to build this one myself.

infispiel: Ah, I see, there is the dockerhub picard image -- it seems like it is build very frequently, too. I suppose the question that remains then is whether it's better to use this automatically updated image or to use a quay image that we know the results of & is stored under HCA.

infispiel: For simplicity's sake we have opted to leave this instance of the change to be on quay. I'm adding a ticket to the backlog to look into updated all of the picard image references in Skylab, and there we will consider whether to use the dockerhub automatic releases or keep releasing our own on quay.

PR #244, opened 2019-08-02
remove the passing build test from the requirements for pullapprove

jessicaway:
### Purpose

- The build test is being split into multiple test entries as portability tests PR #234 . Requiring a passing build test will no longer make sense. We will add the individual tests sback in at a later time. 

### Changes

- Remove the build test from the pullapprove requirements 

### Review Instructions

- No instructions.

### DIFF
--- a/.pullapprove.yml
+++ b/.pullapprove.yml
@@ -1,9 +1,6 @@
 version: 3
 
 pullapprove_conditions:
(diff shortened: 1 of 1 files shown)


jessicaway:
Thanks @rhiananthony !

PR #267, opened 2019-10-09
Remove hard-coded pipeline version from WDLs

samanehsan:
### Purpose
Defining the analysis workflow version in the WDL itself requires keeping that hard-coded version variable up to date. This is prone to manual errors and complicates the release process for pipelines in Skylab.

### Changes
Remove this variable, so that the github tag is the source of truth for the pipeline version

Note: This should be merged after the adapter pipelines are updated to use the pipeline versions specified in Lira's config instead of the pipeline WDLs:
- https://github.com/HumanCellAtlas/adapter-pipelines/pull/6
- https://github.com/HumanCellAtlas/lira/pull/188

### Review Instructions
- No instructions.

### DIFF
--- a/pipelines/cellranger/cellranger.wdl
+++ b/pipelines/cellranger/cellranger.wdl
@@ -2,8 +2,6 @@ workflow CellRanger {
     meta {
         description: "Analyze 3' single-cell RNA-seq data using the 10X Genomics Cellranger pipeline."
     }
-    # version of this pipeline
-    String version = "cellranger_v1.0.2"
 
     String sample_id
     Array[File] fastqs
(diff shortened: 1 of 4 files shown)


samanehsan:
retest this please

samanehsan:
According to @kbergin:
>In our new release process we’ve recently implemented for the rest of the GP Pipelines we also put the version number in the wdl and wrote a script to confirm it’s updated when there have been changes to the pipeline. We intend to output this version number into the headers of bams and similar outputs in the future.

So considering this version could be used in a similar way for the HCA pipelines in the future, we can leave them in for now but still update Lira and the adapter-pipelines to use the pipeline version defined in the config.

PR #268, opened 2019-10-11
Update SmartSeq2 version to v2.5.0

samanehsan:
### Purpose
Since we removed the max retries parameter out of the SmartSeq2 pipeline (in order to allow us to set this value in the workflow options file), we need to tag a new version so that we can make use of this small change.

### Changes
- No changes.

### Review Instructions
- No instructions.


samanehsan:
Agreed! It's hard to look through the commits and figure out which ones affected a particular pipeline.

PR #276, opened 2019-11-01
fix matrix_format_spec 40->80

barkasn:
Trivial fix of documentation of output matrix spec


samanehsan:
Btw it looks like there are merge conflicts with the master branch @barkasn :octocat:

PR #320, opened 2020-04-07
Lk optimus diagram

ekiernan:
### Purpose
Review updated summary diagram

### Changes
Updated the Optimus summary diagram
### Review Instructions

- No instructions.


ekiernan:
Changes requested by NB have been made.

pipelines/optimus/README.md
```diff
@@ -29,7 +29,7 @@
 
 
 # Optimus Pipeline Overview
-![Diagram](https://github.com/HumanCellAtlas/skylab/blob/master/pipelines/optimus/Optimus_diagram.png)
+![Diagram](Optimus_diagram2.png)
```
pshapiro4broad, line 32: Instead of deleting the original and adding a new file, it would be better to replace the original with the new copy. We don't want to create a new file every time we have a new version of the diagram.

ekiernan: I updated the diagram so that it has the original name again.

PR #321, opened 2020-04-07
add optimus 4kpbmc snRNA test

barkasn:
### Purpose

- Add (manually triggered) test that runs the 4kpbmc dataset as a snRNA-seq dataset


test/optimus_snrna/4kpbmc/test_inputs.json
```diff
@@ -0,0 +1,24 @@
+{
+  "TestOptimusPR.expected_bam_hash": "none",
+  "TestOptimusPR.reference_matrix": "gs://hca-dcp-sc-pipelines-test-data/optimus-test-data/referenceMatrix.rds",
+  "TestOptimusPR.expected_cell_metric_hash": "none",
```
jessicaway, line 4: No expected hash?

barkasn: nope! This is just for running the pipeline with this data the output always gets checked manually. reference matrix would be empty too if it didn't break cromwell -- maybe someone knows a way to fix this?

barkasn: Maybe I should actually adjust this so it doesn't run the checker at all -- but actually we do the same thing for the 4kpbmc manual test so maybe that is another ticket.

jessicaway: Maybe we could rename it to indicate that this is a manual test?

PR #322, opened 2020-04-08
add metadata feature to zarr

khajoue2:
### Purpose

- Working on ticket GL-838.

### Changes

- Adding metadata for exon counts.

### Review Instructions

- No instructions.


docker/zarr-output/create_zarr_optimus.py
```diff
@@ -18,13 +18,14 @@
...
 
-def init_zarr(sample_id, path, file_format, schema_version):
+def init_zarr(sample_id, path, file_format, expression_data_type, schema_version):
     """Initializes the zarr output.
 
     Args:
         sample_id (str): sample or cell id
         path (str): path to the zarr output
         file_format (str): zarr file format [DirectoryStore, ZipStore]
+        expression_data_type (str): type of expression data [exonic, whole_trascript]
```
pshapiro4broad, line 28: typo?
```suggestion
        expression_data_type (str): type of expression data [exonic, whole_transcript]
```

library/tasks/ZarrUtils.wdl
```diff
@@ -93,6 +93,8 @@ task OptimusZarrConversion {
   command {
     set -euo pipefail
 
+    [[ ~{counting_mode} == "sc_rna" ]] && expression_data_type_param="exonic"  || expression_data_type_param="whole_transcript"
```
pshapiro4broad, line 96: This code doesn't seem complete

library/tasks/ZarrUtils.wdl
```diff
@@ -103,7 +105,8 @@ task OptimusZarrConversion {
        --output_path_for_zarr "${sample_id}.zarr" \
        --format DirectoryStore \
        --sample_id ${sample_id} \
-       --count_matrix ${sparse_count_matrix}
+       --count_matrix ${sparse_count_matrix} \
+       --expression_data_type $expression_data_type_param
```
pshapiro4broad, line 109: You need `{}` to expand a WDL input. For all of these cases I would also use `~{}` instead of `${}`, as `~{}` is the newer form and doesn't cause conflicts with bash's `${}` expression.
```suggestion
       --expression_data_type ~{expression_data_type_param}
```

When this code was converted to WDL 1.0, the $'s should've been changed to ~'s but we must've missed it then.

library/tasks/ZarrUtils.wdl
barkasn: It acts like a ternary operator in C

library/tasks/ZarrUtils.wdl
barkasn: Yes, this needs fixing

barkasn: we actually have a lot of those and ticket to change them

library/tasks/ZarrUtils.wdl
barkasn: Maybe we should expand it to an if for readability?

library/tasks/ZarrUtils.wdl
barkasn: Actually no, I am wrong the ${} need to be changed to ~{} and we have another ticket. But $expression_data_type_param is fine because its a bash variable set in the script.

library/tasks/ZarrUtils.wdl
pshapiro4broad: But `expression_data_type_param` has no value here, other than the text itself. It's not declared as a task input.

library/tasks/ZarrUtils.wdl
pshapiro4broad: > Actually no, I am wrong the ${} need to be changed to ~{} and we have another ticket. But $expression_data_type_param is fine because its a bash variable set in the script.

Ohhh.. yeah that is very confusing. Could you instead use an `if`? and make the bash variable all uppercase, as is the usual convention for bash variables. That would help readability for me, at least.

khajoue2: I just made the changes.

PR #324, opened 2020-04-15
Fk gl 839

khajoue2:
### Purpose
Add expression_data_type to Optimus Loom Output

### Changes

 Added global attribute "expression_data_type" to the loom output

### Review Instructions

- No instructions.


docker/zarr-output/create_zarr_optimus.py
```diff
@@ -463,14 +464,15 @@ def create_zarr_files(args):
     Args:
         args (argparse.Namespace): input arguments for the run
     """
-    version = "1.0.0"
+    version = "1.0.1"
 
     # initiate the zarr file
     root_group = init_zarr(
```
barkasn, line 470: I think it would be good practice to convert all these to named arguments as opposed to positional

docker/zarr-to-loom/optimus_zarr_to_loom.py
```diff
@@ -37,13 +35,16 @@ def main():
...
     root = zarr.open(store)
-
+    
+    #Get expression data type: exonic or whole_transcript
+    expression_data_type = root.attrs['expression_data_type']
+    
     # Get the expression matrix
     # expression matrix in numpy ndarray format (dense)
     # NOTE: If memory is limiting this could be done by chunk
-    nrows = root[f"/{sample_id}.zarr/expression"].shape[0]
```
barkasn, line 44: Although good simplifications these changes may be assuming different input and or otherwise be breaking the pipeline. Please investigate why the tests failed.

library/tasks/ZarrUtils.wdl
```diff
@@ -93,6 +93,8 @@ task OptimusZarrConversion {
   command {
     set -euo pipefail
 
+    [[ ~{counting_mode} == "sc_rna" ]] && EXPRESSION_DATA_TYPE_PARAM="exonic"  || EXPRESSION_DATA_TYPE_PARAM="whole_transcript"
```
barkasn, line 96: I thought that this was going to get expanded to an if for readability

docker/zarr-to-loom/optimus_zarr_to_loom.py
jessicaway: The current error message on the failed Optimus run is:
 `Failed to process 'call ZarrUtils.OptimusZarrToLoom' (reason 1 of 1): The call supplied a value 'sample_id' that doesn't exist in the task (or sub-workflow))`

PR #327, opened 2020-04-21
Input array of read1 and read2 instead of assuming a particular file naming scheme

joshua-gould:
### Purpose

- No issue is linked to this PR.

### Changes

- No changes.

### Review Instructions

- No instructions.


nikellepetrillo:
@joshua-gould we have implemented a similar change for this workflow in Terra. We are looking at how we want to address this issue for the long term. We will make these changes in the WARP repo https://github.com/broadinstitute/warp

PR #329, opened 2020-04-24
Update Optimus.changelog.md

ekiernan:
### Purpose
Review updated Optimus Changelog
### Changes

- No changes.

### Review Instructions

- No instructions.

### DIFF
--- a/pipelines/optimus/Optimus.changelog.md
+++ b/pipelines/optimus/Optimus.changelog.md
@@ -1,4 +1,20 @@
+# (next version)
+2020-04-24 (Date of Last Commit)
+
+* Added new snRNAseq counting_mode parameter
+  * For the sn_rna couting mode only, updated dropseq tools to version #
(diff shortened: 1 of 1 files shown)


pipelines/optimus/Optimus.changelog.md
```diff
@@ -1,4 +1,24 @@
+# (next version)
+2020-04-24 (Date of Last Commit)
+
+* Added a new counting_mode parameter to Optimus workflow which enables processing of single-nuclei datasets
+
```
gbggrant, line 5: Is the standard to add a blank line between each bulleted item in the changelog?  I don't think we're doing that in dsde-pipelines.

ekiernan: We never talked about when making the style guide- I will have to update the guide. I changed the changelog here to be single space so that it matches dsde-pipelines.

PR #337, opened 2020-05-07
Kmk mt metric

kishorikonwar:
### Purpose
To add mitochondrial gene counts to the cell metrics

### Changes
 Changed the CalcualteCellMetrics input to add the reference gtf files and the docker for sctools in the same task


library/tasks/SequenceDataWithMoleculeTagMetrics.wdl
```diff
@@ -47,9 +47,10 @@ task CalculateGeneMetrics {
 task CalculateCellMetrics {
   input {
     File bam_input
+    File original_gtf
 
     # runtime values
-    String docker = "quay.io/humancellatlas/secondary-analysis-sctools:v0.3.7"
+    String docker = "quay.io/humancellatlas/secondary-analysis-sctools:kmk-mt-counts"
```
jessicaway, line 53: can you rename the docker with a version number?

kishorikonwar:
retest

library/tasks/SequenceDataWithMoleculeTagMetrics.wdl
kishorikonwar: Yes, the version number is changed.

library/tasks/ZarrUtils.wdl
```diff
@@ -93,12 +93,12 @@ task OptimusZarrConversion {
   command {
     set -euo pipefail
 
-
-    if ~{counting_mode} == "sc_rna"
-    then
+    if [ "${counting_mode}" == "sc_rna" ]; then
         EXPRESSION_DATA_TYPE_PARAM="exonic" 
+        ADD_EMPTYDROPS_DATA="yes"
```
jessicaway, line 98: Can we use boolean values instead of strings here?

kishorikonwar: We could but the python script, for which this variable is prepared, requires strings as argument options.

PR #346, opened 2020-05-28
Remove Zarr from ss2 and optimus pipeline GL-1008

khajoue2:
### Purpose
Removing zarr files from Optimus pipeline and generating loom files directly.

- No issue is linked to this PR.

### Changes

- No changes.

### Review Instructions

- No instructions.


library/tasks/SequenceDataWithMoleculeTagMetrics.wdl
```diff
@@ -50,7 +50,7 @@ task CalculateCellMetrics {
     File original_gtf
 
     # runtime values
-    String docker = "quay.io/humancellatlas/secondary-analysis-sctools:v0.3.8"
+    String docker = "quay.io/humancellatlas/secondary-analysis-sctools:kmk-mt-counts"
```
kbergin, line 53: We should be using a versioned release of docker images, not one with a branch name in it. Can this be updated before release?

pipelines/optimus/Optimus.changelog.md
```diff
@@ -1,3 +1,9 @@
+#optimus_v2.1.0
+
+* Removed zarr file
+
+* Added emptyDrops for sn_rna mode
```
kbergin, line 5: Didn't we actually remove emptyDrops, not add it? 

I think we need to add a bit more information about the removal of zarr.

Something like "Removed zarr formatted matrix and metrics outputs and replaced with loom" and I believe this is a breaking change. If this is going out with the features for the snRNAseq support it's definitely a major version update. Are the snRNAseq changes included in this PR?

PR #352, opened 2020-06-09
changed merged_bam output file name to sample_id.bam

khajoue2:
### Purpose
Every bam made by Optimus is called merged.bam. From the DCP perspective, the infrastructure will require the outputs of pipelines to have a unique name for the entire HCA. From a user's perspective, they want certain names for their files. 

### Changes

- No changes.

### Review Instructions

- No instructions.

### DIFF
--- a/pipelines/optimus/Optimus.changelog.md
+++ b/pipelines/optimus/Optimus.changelog.md
@@ -1,4 +1,4 @@
-#optimus_v3.0.0
+# optimus_v3.0.0
 
 * Removed zarr formatted matrix and metrics outputs and replaced with loom
 
--- a/pipelines/optimus/Optimus.wdl
+++ b/pipelines/optimus/Optimus.wdl
@@ -222,6 +222,7 @@ workflow Optimus {
   call Merge.MergeSortBamFiles as MergeSorted {
     input:
       bam_inputs = PreMergeSort.bam_output,
+      sample_id = sample_id,
       sort_order = "coordinate"
   }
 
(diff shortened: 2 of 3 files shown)


kbergin:
We discussed this in refinement - did we consider this for the implementation? Or are we thinking users will override the `sample_id` input with whatever they want? 

> We could use sample_id as default naming of pipeline outputs but give user an option to override this as an optional inputWe could use sample_id as default naming of pipeline outputs but give user an option to override this as an optional input

pipelines/optimus/Optimus.changelog.md
```diff
@@ -1,4 +1,4 @@
-#optimus_v3.0.0
+# optimus_v3.0.0
 
 * Removed zarr formatted matrix and metrics outputs and replaced with loom
```
ekiernan, line 3: Could we add another bullet point * "Added metrics for mitochondrial reads"

pipelines/optimus/Optimus.wdl
```diff
@@ -222,6 +222,7 @@ workflow Optimus {
   call Merge.MergeSortBamFiles as MergeSorted {
     input:
       bam_inputs = PreMergeSort.bam_output,
+      sample_id = sample_id,
```
pshapiro4broad, line 225: I think it might be easier to write if you change `MergeSortBamFiles` to accept a `String` parameter `output_bam_filename`. That way the task can use `output_bam_filename` throughout.

pipelines/optimus/Optimus.changelog.md
khajoue2: Done.

pipelines/optimus/Optimus.wdl
khajoue2: That makes sense, I changed it to take output_bam_filename from Optimus wdl

PR #358, opened 2020-07-28
increased the HDD to 3500 GB, from 500 GB originally, to speed-up the SplitBamByCellBarcode step

kishorikonwar:
### Purpose
Speed-up the SplitCellByCellBarcode step of Optimus by modifying the sctools to write the intermediate BAM files of SplitCellByBarcode in SAM format. The goal is to speed-up the step but at the expense of larger sized intermediate files in the step

### Changes
a) Changed the HDD from 500GB to 3500GB 
b) Changed the docker image to a higher version of sctools


nikellepetrillo (approved):
looks good as long as it's okay to use 3500GB for disk!

library/tasks/SplitBamByCellBarcode.wdl
```diff
@@ -6,13 +6,14 @@ task SplitBamByCellBarcode {
...
     # runtime values
-    String docker = "quay.io/humancellatlas/secondary-analysis-sctools:v0.3.5"
+    String docker = "quay.io/humancellatlas/secondary-analysis-sctools:v0.3.9"
 
     Int machine_mem_mb = 15258
     Int cpu = 16
 
     # we can calculate disk size for arrays of input files in WDL 1.0
-    Int disk = 500
+    Int disk = 3500
```
jessicaway, line 15: should we calculate this based on input size rather than just always getting a ton of disk?

